PR #5, opened 2026-02-10
feat(auth): add stitch-auth package  (STIT-333, STIT-335)

mbarlow12:
## Summary
- Adds the `stitch-auth` package (`packages/stitch-auth/`) for OIDC JWT token validation
- Registers `stitch-auth` as a workspace member in the root `pyproject.toml`
- Includes unit tests (28 tests) for claims, settings, and validator

## Test plan
- [x] `uv run --package stitch-auth --exact pytest packages/stitch-auth/tests/ -xvs` — 28 passed
- [x] `ruff check` — all checks passed

### DIFF
--- a/packages/stitch-auth/src/stitch/auth/validator.py
+++ b/packages/stitch-auth/src/stitch/auth/validator.py
@@ -34,7 +34,7 @@ class JWTValidator:
                 issuer=self._settings.issuer,
                 leeway=timedelta(seconds=self._settings.clock_skew_seconds),
                 options={
-                    "require": ["exp", "iss", "aud", "sub", "nbf"],
+                    "require": ["exp", "iss", "aud", "sub"],
                     "verify_exp": True,
                     "verify_iss": True,
                     "verify_aud": True,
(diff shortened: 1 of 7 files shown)


packages/stitch-auth/src/stitch/auth/validator.py
```diff
@@ -0,0 +1,56 @@
...
+        except jwt.ExpiredSignatureError as e:
+            raise TokenExpiredError(str(e)) from e
+        except jwt.InvalidTokenError as e:
+            raise TokenValidationError(str(e)) from e
+
+        email = payload.get("email") or payload.get("preferred_username")
+        name = payload.get("name")
+
+        return TokenClaims(
+            sub=payload["sub"],
```
AlexAxthelm: Maybe blocking: related to discussion in #6, should issuer be part of the token claims here? so that an identical sub from a different issuer can't spoof a claim?

AlexAxthelm (approved):
Approved, pending resolution of open comment.
cc @jdhoffa

PR #19, opened 2026-02-24
feat(workspace): add `stitch-ogsi` package (STIT-360)

mbarlow12:
Adds OGSI-specific models, types, and constants.

Since there's not really any logic here that's not built into `pydantic`, I think a lot of the tests we might want to see are reflected in #18.

### DIFF
--- a/packages/stitch-ogsi/src/stitch/ogsi/model/og_field.py
+++ b/packages/stitch-ogsi/src/stitch/ogsi/model/og_field.py
@@ -67,10 +67,10 @@ class OilGasFieldBase(BaseModel):
     basin: str | None = None
     """Geological basin name."""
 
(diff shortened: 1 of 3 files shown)


AlexAxthelm (dismissed):

packages/stitch-ogsi/src/stitch/ogsi/model/og_field.py
```diff
@@ -0,0 +1,98 @@
...
+
+from .types import (
+    LocationType,
+    ProductionConventionality,
+    PrimaryHydrocarbonGroup,
+    FieldStatus,
+)
+
+
+class Owner(BaseModel):
```
AlexAxthelm:
```suggestion
class OilGasOwner(BaseModel):
```

PR #21, opened 2026-02-27
refactor: remove unused `stitch-core` package

AlexAxthelm:
Removes the stitch-core package, which is not currently used in other parts of the project, and core logic has been translated to either the `deployments/api` package, or data models in #18, #19


jdhoffa (approved):
Good with me, but defer to @mbarlow12 in case there is anything worth keeping from there

AlexAxthelm:
Merging, since this is `git`, and anything we want to pull out of this can be restored from old commits.

PR #53, opened 2026-03-31
entity linkage (STIT-342)

AlexAxthelm:
Introduce the first of the stitch "friends": Entity-linkage.

This is still a proof-of-concept (intentionally simple logic), but captures the flow of a more complex process:

* Query the paginated list endpoint to accumulate an in-memory representation of resources sufficient to identify potential matches.
* for each potential match: get the "detail" info, and confirm match.
* optionally: post to the `merge` endpoint

Stands as an independent FastAPI service, and includes a minimal frontend for demo purposes.

Copies parts of the `api` service as a framework, then some alterations to simplify. Frontend section written by chatGPT.

Auth Note: This currently passes through the user's token, rather than trying machine auth.

confirmed locally with all 3 main entrypoints in Makefile


deployments/entity-linkage/src/stitch/entity_linkage/client.py
```diff
@@ -0,0 +1,153 @@
...
+        self._raise_for_status(response, f"GET /oil-gas-fields/{resource_id}/detail")
+        payload = response.json()
+        data = payload.get("data") or {}
+        return FieldDetailCandidate(
+            id=payload["id"],
+            name=data.get("name"),
+            country=data.get("country"),
+        )
+
+    async def post_merge(
```
AlexAxthelm, line 134: @jdhoffa The client for `seed` is set up a little differently from this one, but it looks very translatable to this structure (additions to this client). 

For an llm service, I would use the mechanisms here to get resource details, then generate a set of resources (`iterator`, see `deployments/seed/src/stitch/seed/payloads.py`), and post them (payloads again, also `deployments/seed/src/stitch/seed/openapi_validate.py` to get validation before post). Then this branch has the mechanisms to add those new resources to the merge review pile, and boom, that's the llm review flow.

jdhoffa (changes requested):
This PR ships a new service that can perform destructive merges, but it currently doesn't have any automated test coverage.

For a new service that both decides merge candidates and can apply them, I’d block merge until there is at least coverage for the matching logic and the `/start` happy-path / failure-path behavior.

jdhoffa (approved):

PR #56, opened 2026-04-01
Stit 406 frontend explorer list page

jasonbrownrmi:
- **Pagination**: adds a `Pagination` component and threads `page` / `page_size` as query params through `useResources` → `getResources`. Page state is preserved in `searchParams` so navigating back from a detail page returns to the correct page.
- **Filtering**: `FilterBar` options now correctly read from `resource.data[field]` (fixing a broken accessor). Active filters are sent to the API as repeated query params (e.g. `?basin=Arabian&basin=Permian`), replacing the previous client-side filter. Presently the API handles a single str for each filter. We might consider expanding this to handle a list.
- **Sorting**: `sortConfig` is lifted from `ResourcesTable` into `ResourcesView`. Clicking a column header sends `sort_by` / `sort_order` to the API; the table renders sort indicators from the lifted state. Client-side sort is removed.
- **Tests**: `ResourcesView.test.jsx` and `api.test.js` updated to cover pagination params, filter params as repeated query params, sort param forwarding, and filter/sort reset on cache clear.


AlexAxthelm (approved):
Looks good! Some weirdness with filter options and counts, but fixing those is out of scope here (api changes).

Tested:

* filtering
* sorting

PR #58, opened 2026-04-07
Feat/merge candidates

AlexAxthelm:
Introduce a merge-candidate workflow for oil & gas fields, replacing immediate merges with a review step. Adds backend support for creating, approving/denying, and previewing merge candidates (cc @mbarlow12), plus corresponding frontend review UI (cc @jasonbrownrmi).

Also updates entity-linkage to submit merge candidates instead of applying merges directly, aligning all flows with the new review-based process.

Accessible through the api's `oil-gas-field/merge-candidates` family of endpoints

Fun review/demo:
1. Start system with `seed` and `entity-linkage` active (any of the entrypoints like `make reboot-docker`)
2. Go to "Entity Linkage" at the top bar, run with "Apply merges"
3. go to "Merge Review"
4. See the merges to review, approve, deny, etc


jdhoffa (changes requested):
Approved candidates break in the review page.

The UI always requests `GET /merge-candidates/{id}/preview`, but once a candidate is approved its resources have already been merged, and `preview_merge_candidate()` rejects them as non-mergeable.

Result: select a pending candidate, preview works; approve it; select it again, preview errors.

Since the page explicitly shows approved candidates, this leaves the reviewed state broken.

Repro:
1. Open Merge Candidate Review.
2. Select a PENDING candidate and confirm that preview loads.
3. Approve it.
4. Re-select that same candidate.
5. Preview now errors.

A possible fix:
In the frontend: only fetch preview for `PENDING` candidates. For `APPROVED` and `DENIED` candidates, render the candidate metadata and resources without trying to regenerate the preview.

AlexAxthelm:
@jdhoffa good catch. The current version shows an error message there on merge, but moving to a different candidate and then back to the not-PENDING one doesn't show preview. I'm not good enough at JS to solve further.

jdhoffa:
> @jdhoffa good catch. The current version shows an error message there on merge, but moving to a different candidate and then back to the not-PENDING one doesn't show preview. I'm not good enough at JS to solve further.

I'm good with this! thanks

jdhoffa (approved):

PR #60, opened 2026-04-09
Add preview-only AI field suggestions to resource detail

jdhoffa:
This PR adds a (very) small MVP for AI-generated field suggestions on the oil and gas resource detail pages. 
Note: There IS NOT persistance layer at this stage. The suggestion is totally ephemoral and rendered on the front-end. 
The priority here was to focus on wiring up Azure Foundry and getting the suggestion calls wired up. Persistance will be handled in a future PR. 

Users can now open an existing resource, choose a "supported" field, and request an AI suggestion, and preview that suggested value in the detail UI. 

Next steps:
The current implementation is not doing any grounded web retrieval, so returning URLs at this stage would be misleading (basically just a hallucination).  We are going to need to figure out how to connect the model to the Web to parse public sources for citation URLs... otherwise we are just relying on hallucinations here

## Changes to ENV
Note, the following environment variables are added in this PR. They are available in LastPass:
``` bash
AZURE_OPENAI_API_KEY=...
AZURE_OPENAI_ENDPOINT=https://<your-resource>.openai.azure.com
AZURE_OPENAI_DEPLOYMENT=<your-deployment-name>
AZURE_OPENAI_API_VERSION=2024-10-21
```


jdhoffa:
Screenshot 2026-04-09 at 13 08 21

deployments/api/src/stitch/api/llm_suggestions.py
```diff
@@ -0,0 +1,175 @@
...
+    "production_start_year",
+    "location_type",
+    "field_status",
+    "primary_hydrocarbon_group",
+    "production_conventionality",
+)
+
+
+class LLMFieldSuggestionRequest(BaseModel):
+    field: str
```
mbarlow12, line 26:
```suggestion
type LLMAllowedSuggestionFields = Literal[
    "basin",
    "state_province",
    "discovery_year",
    "fid_year",
    "production_start_year",
    "location_type",
    "field_status",
    "primary_hydrocarbon_group",
    "production_conventionality",
]


class LLMFieldSuggestionRequest(BaseModel):
    field: LLMAllowedSuggestionFields
```
This gives us built-in validation from Pydantic and could remove `validate_llm_suggestion_field`.

env.example
```diff
@@ -8,6 +8,10 @@ ENTITY_LINKAGE_LOG_LEVEL=info
 
 # API
 AUTH_DISABLED=true
+AZURE_OPENAI_API_KEY=replace-me
```
AlexAxthelm, line 11: NB: These are different var names than are in LastPass

deployments/stitch-frontend/src/constants/fieldMeta.js
```diff
@@ -45,3 +45,15 @@ export const IDENTITY_FIELDS = Object.entries(FIELD_META)
 export const PRODUCTION_FIELDS = Object.entries(FIELD_META)
   .filter(([, v]) => v.section === "production")
   .map(([k]) => k);
+
+export const AI_SUGGESTION_FIELDS = [
```
AlexAxthelm, line 49: Open question: Do we want this list to only show unpopulated fields in the frontend (don't generate suggestions for existing fields)?7

deployments/api/src/stitch/api/llm_suggestions.py
```diff
@@ -0,0 +1,175 @@
...
+
+import httpx
+from pydantic import BaseModel, ConfigDict, ValidationError
+
+from stitch.api.settings import Settings
+from stitch.ogsi.model import OGFieldDetailView
+from stitch.ogsi.model.og_field import OilGasFieldBase
+
+
+ALLOWED_LLM_SUGGESTION_FIELDS = (
```
AlexAxthelm, line 12: NB: Seems like there's chance of drift between this definition of the permitted fields and the one in frontend. Maybe there's a good way to write once and read where needed (@mbarlow12 ?)

deployments/api/src/stitch/api/llm_suggestions.py
```diff
@@ -0,0 +1,175 @@
...
+        },
+    ]
+
+
+def parse_llm_suggestion_response(
+    raw_response: str, *, requested_field: str
+) -> ParsedLLMFieldSuggestion:
+    try:
+        payload = json.loads(_strip_code_fences(raw_response))
+        parsed = ParsedLLMFieldSuggestion.model_validate(payload)
```
AlexAxthelm, line 103: Noting that this model validation is distinct (unrelated) to the OGSIModel validation, which could give us a value that is LLM-valid, but not consistent with the rest of the data model. 

Not a big concern at this point, when we aren't persisting anything, but sooner than later on passing it through a "real" validation.

deployments/api/src/stitch/api/routers/oil_gas_fields.py
```diff
@@ -67,6 +77,57 @@ async def get_resource_detail(
     return resource_to_detail_view(resource=res)
 
 
+@router.post("/{id}/llm-suggestions", response_model=LLMFieldSuggestionResponse)
```
AlexAxthelm, line 231: NB: future discussion, how much of the "stitch friends" (Entity Linkage, LLM backfill) do we want in the core API contract?

AlexAxthelm (approved):
Looks good! Some points to discuss for future work, but does what it says on the tin.
 
* Successfully displays "LLM suggestions are not configured." when envvars not set
* Returns no suggestion when minimal information
* returns actual value whe requested to fill existing populated field

jdhoffa:
After a chat with @AlexAxthelm we determined that building LLM context directly using the internal `resource_to_detail_view` function probably isn't the right way to do this.
We need to ensure that the LLM's access permissions are appropriately scoped and the best/ easiest way to do that is using the same infrastructure that we will be using for user tokens 

Likely, a better way to solve this will involve building an LLM service which accesses the Stitch API service rather than directly accessing internal `db` functions.

@AlexAxthelm will fork this branch and work to extract the functionality as it's own service.

jdhoffa:
Converting to draft for now, to preserve this functionality as it does work.
Will close the PR once we have a suitable replacement.

env.example
jdhoffa: Updated

deployments/api/src/stitch/api/llm_suggestions.py
```diff
@@ -0,0 +1,175 @@
...
+        assert deployment is not None
+        assert api_version is not None
+
+        self._api_key = api_key.get_secret_value()
+        self._endpoint = str(endpoint).rstrip("/")
+        self._deployment = deployment
+        self._api_version = api_version
+
+    async def generate_field_suggestion(self, *, messages: list[dict[str, str]]) -> str:
+        url = f"{self._endpoint}/openai/deployments/{self._deployment}/chat/completions"
```
mbarlow12, line 146: NB: Not sure if this is intentional, but looks like the current recommendation is the [`responses` API](https://developers.openai.com/api/reference/responses/overview). Looks like it's supported via Azure as well.

We'd probably get some mileage out of the `openai` python sdk as well...maybe.

deployments/api/src/stitch/api/llm_suggestions.py
```diff
@@ -0,0 +1,175 @@
...
+        params = {"api-version": self._api_version}
+        headers = {"api-key": self._api_key, "Content-Type": "application/json"}
+        payload = {
+            "messages": messages,
+            "response_format": {"type": "json_object"},
```
mbarlow12, line 151: You can also pass a schema here, something like
```python
"response_format": ParsedLLMFieldSuggestion.model_json_schema(),
```
supported in both chat completions and responses api. Might help in easier response handling.
[completions reference](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#(resource)%20chat.completions%20%3E%20(method)%20create%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20response_format%20%3E%20(schema))

mbarlow12 (approved):
👍

jdhoffa:
Likely superseded by #73

AlexAxthelm:
@jdhoffa vote to close as superseded by #73

PR #63, opened 2026-04-17
Read config from JSON file, rather than Envvars

AlexAxthelm:
Switch the frontend from build-time browser env injection to runtime public/config.json, so the bundle can independently of knowing values for depoy environment (local, preview, and production). This keeps build metadata at build time while moving API/Auth/service endpoints to deploy-time config.

The PR also cleans up related local-dev wiring: removes stale frontend Docker/build args, updates docs, and aligns api-dev/entity-linkage behavior with the new config model. Tests were updated to initialize runtime config and match the new bootstrap flow.

Tested to work with all 3 `make` devenvs


deployments/stitch-frontend/src/config/env.js
```diff
@@ -33,9 +23,64 @@ function loadConfig() {
...
+      "Config has not been initialized yet. Call loadConfig() during app bootstrap before importing modules that use config.",
+    );
+  }
+  return resolvedConfig;
+}
+
+export function setConfigForTests(config) {
+  resolvedConfig = Object.freeze(config);
+  configPromise = Promise.resolve(resolvedConfig);
+}
```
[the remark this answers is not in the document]
AlexAxthelm: Willing to accept this in tests

jdhoffa:
Hey @AlexAxthelm, just checking if you have coordinated this frontend change with @jasonbrownrmi as it will require changes to his local development workflow? 

I don't have any strong fundamental opinion for why we shouldn't use a `.json` instead of a `.env` (beyond it being something I'm not super used to), but the key reviewer and approver here will need to be @jasonbrownrmi

jdhoffa (changes requested):
Review comments created by `codex`, reviewed and edited by me:

Blocking:
This PR switches frontend runtime config from build-time `VITE_*` vars to `/config.json`, but `node-build` still computes `VITE_APP_ENV` as if it controls runtime behavior, while the `Dockerfile` now just copies the committed `public/config.json`. For predictable behavior, we should have one source of truth: either remove the stale `VITE_APP_ENV` path here, or add an explicit step that generates the runtime `config.json` for each build.

https://github.com/RMI/stitch/blob/921fb74813f0f803db7b6169759f64961bac6892/.github/workflows/node-build.yml#L20-L50

---

Non-blocking:
The new runtime-config model depends on `/config.json` being non-cacheable, but the nginx config in this repo does not enforce that. The docs explicitly require `Cache-Control: no-store` (`deployments/stitch-frontend/README.md:164`), yet `deployments/stitch-frontend/nginx.conf:1` only special-cases hashed assets and leaves `/config.json` on default caching behavior. That can leave clients pinned to stale backend/Auth0 settings across redeploys.

Non-blocking:
`make frontend-test passes`, but the updated `ColophonPanel` tests now emit React `act(...)` warnings during the suite (`deployments/stitch-frontend/src/components/ColophonPanel.test.jsx:196`, `deployments/stitch-frontend/src/components/ColophonPanel.test.jsx:207`). Not a release blocker, but it adds noise and can hide real async test regressions.

deployments/stitch-frontend/nginx.conf
```diff
@@ -10,11 +10,16 @@ server {
     try_files $uri $uri/ /index.html;
   }
 
+  # Runtime config must not be cached across deployments.
+  location = /config.json {
+    try_files $uri =404;
+    add_header Cache-Control "no-store";
```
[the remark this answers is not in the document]
AlexAxthelm: this can be addressed when we look closer at our local nginx usage

deployments/stitch-frontend/src/config/env.js
```diff
@@ -1,29 +1,19 @@
...
+      domain: runtimeConfig.auth0Domain,
+      clientId: runtimeConfig.auth0ClientId,
+      audience: runtimeConfig.auth0Audience,
     }),
-    apiBaseUrl: import.meta.env.VITE_API_URL || "http://localhost:8000/api/v1",
-    appEnv:
-      import.meta.env.VITE_APP_ENV || import.meta.env.MODE || "development",
-
+    apiBaseUrl: runtimeConfig.apiUrl || "http://localhost:8000/api/v1",
+    appEnv: runtimeConfig.appEnv || "development",
```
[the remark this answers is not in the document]
AlexAxthelm: This is mandatory in config, and run should fail if missing.

deployments/stitch-frontend/src/components/ColophonPanel.jsx
```diff
@@ -35,6 +35,7 @@ function useSystemInfo() {
 }
 
 function formatBackendSection(state) {
```
jasonbrownrmi: calls getConfig() every invocation. It works because the function is only ever called during render (after bootstrap), but it's a footgun for whoever adds the next helper — someone could easily call one of these from a top-level import side effect and get the "Config has not been initialized" throw. Consider a lint rule, or passing config as a parameter to pure helpers so the "runtime only" contract is explicit.

deployments/stitch-frontend/src/components/ColophonPanel.jsx
```diff
@@ -163,29 +165,27 @@ export default function ColophonPanel({ diagnosticsOpen = false }) {
     };
   }, [getAccessTokenSilently, isAuthenticated, isLoading]);
 
-  const sections = useMemo(() => {
```
jasonbrownrmi, line 166: is there a reason why useMemo was dropped here? It should work either way but wonder if there was a reason for it.

deployments/stitch-frontend/src/main.jsx
```diff
@@ -18,24 +18,41 @@ const queryClient = new QueryClient({
...
-          </BrowserRouter>
-        </AuthGate>
-      </QueryClientProvider>
-    </Auth0Provider>
-  </StrictMode>,
-);
+async function bootstrap() {
+  const config = await loadConfig();
+
+  createRoot(document.getElementById("root")).render(
```
jasonbrownrmi: createRoot doesn't run until await loadConfig() resolves. On a slow network the user might see a white page until the fetch completes, and only a plain Configuration error if it fails. A static loading shell in index.html could cover both cases cheaply.

Something like this in index.html:                                                                                                 
                                                                                                                                     
  ```<body>                                                                                                                             
    <div id="root">
      <div class="bootstrap-shell" role="status" aria-live="polite">
        <div class="bootstrap-shell__spinner" aria-hidden="true"></div>                                                              
        <p class="bootstrap-shell__label">Loading Stitch…</p>                                                                        
      </div>                                                                                                                         
    </div>                                                                                                                           
    <style>       
      .bootstrap-shell {                                                                                                             
        min-height: 100vh;
        display: flex;
        flex-direction: column;                                                                                                      
        align-items: center;
        justify-content: center;                                                                                                     
        gap: 0.75rem;
        font-family: system-ui, -apple-system, sans-serif;
        color: #475569;                                                                                                              
      }
      .bootstrap-shell__spinner {                                                                                                    
        width: 2rem;                                                                                                                 
        height: 2rem;
        border: 2px solid #e2e8f0;                                                                                                   
        border-top-color: #475569;
        border-radius: 50%;                                                                                                          
        animation: bootstrap-spin 0.8s linear infinite;
      }                                                                                                                              
      .bootstrap-shell__label { font-size: 0.875rem; margin: 0; }
      @keyframes bootstrap-spin { to { transform: rotate(360deg); } }                                                                
    </style>                                                                                                                         
    <script type="module" src="/src/main.jsx"></script>                                                                              
  </body>                                            ```

.github/workflows/node-build.yml
```diff
@@ -26,14 +26,6 @@ jobs:
           NODE_VERSION=$(node -p "process.version")
           VITE_VERSION=$(node -p "require('./deployments/stitch-frontend/package.json').devDependencies.vite")
```
jasonbrownrmi, line 28: Is there a build time mechanism to add the config.json? What config.json does a deployed instance serve?

deployments/stitch-frontend/src/queries/api.js
```diff
@@ -1,10 +1,11 @@
-import config from "../config/env";
+import { getConfig } from "../config/env";
 
 export async function getResources(
   fetcher,
   endpoint = "resources",
   { page = 1, page_size = 50, filters = {}, sort_by, sort_order } = {},
 ) {
+  const config = getConfig();
```
jasonbrownrmi: Lots of repetition with this getConfig() call. It works fine, but it might be worth mentioning the reason for calling it within each function (maybe a comment in env.js) to discourage attempts to use at a module scope.

jasonbrownrmi (commented):
I left comments in the code for your review. Nothing particularly blocking.

deployments/stitch-frontend/src/main.jsx
AlexAxthelm: https://github.com/RMI/stitch/pull/63/commits/72a1c6b15c6d9f448597629daaef1e923d4f0789

Good catch. it all works great on dev/test because there's no network

@jdhoffa STIT-449 added to tech debt list to simulate network failures/latency

deployments/stitch-frontend/src/components/ColophonPanel.jsx
AlexAxthelm: Thanks for pointing this out @jasonbrownrmi. I'm still getting used to the idea that a SPA may load many files.

The bootstrap process is now explicit (with loading placeholder shell) and loads a `ConfigContext`, which is getting called as a reference with `useConfig` (rather than loaded with `loadConfig`) or passed as arguments. There's still a lot of `useConfig` around to make it explicit when we're leaning on that singleton, but those are all replacing the previous reference to the built `config` object.

deployments/stitch-frontend/src/queries/api.js
AlexAxthelm: refactored to use a Config Context

deployments/stitch-frontend/src/components/ColophonPanel.jsx
AlexAxthelm: This is pretty cheap to calculate, and this codepath only runs in non-PROD environment

deployments/stitch-frontend/src/config/env.js
```diff
@@ -1,29 +1,50 @@
...
+function freezeConfig(runtimeConfig) {
   return Object.freeze({
     auth0: Object.freeze({
-      domain: import.meta.env.VITE_AUTH0_DOMAIN,
-      clientId: import.meta.env.VITE_AUTH0_CLIENT_ID,
-      audience: import.meta.env.VITE_AUTH0_AUDIENCE,
+      domain: runtimeConfig.auth0Domain,
+      clientId: runtimeConfig.auth0ClientId,
+      audience: runtimeConfig.auth0Audience,
     }),
```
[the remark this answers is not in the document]
AlexAxthelm: handled in sample local config

deployments/stitch-frontend/src/main.jsx
```diff
@@ -18,24 +19,43 @@ const queryClient = new QueryClient({
...
+    <StrictMode>
+      <ConfigProvider config={config}>
```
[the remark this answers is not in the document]
AlexAxthelm: https://github.com/RMI/stitch/pull/63/commits/d95dda7276dfe35c3fbdbf01949291162f5e6ddb

.github/workflows/node-build.yml
AlexAxthelm: @jasonbrownrmi This is added in https://github.com/RMI/stitch/pull/70, but basically the build process includes the `config.json` that's in `public/` (dev), as part of `dist/`, but overwriting `dist/config.json` gives you the ability to change what we need for the deployment

jasonbrownrmi (approved):

jdhoffa (approved):

PR #66, opened 2026-04-17
prepare db deployment for CD

AlexAxthelm:
Changes to DB deployment discovered while preparing CD pipeline:

do not fail on existing roles, and update docs to allow alternate db names


jdhoffa (approved):
LGTM

PR #68, opened 2026-04-20
Ci/deploy pipeline

AlexAxthelm:
Adds a GitHub Actions CD pipeline for Stitch preview and main deployments.

This PR introduces reusable workflows to:

* Build and publish Docker images for api, entity-linkage, and seed (lightly modified from PACTA workflows)
* Provision and initialize a Postgres database for preview/main environments
* Deploy the API and entity-linkage services to Azure Container Apps
* Run DB init and seed jobs from GitHub Actions
* Tear down preview resources when PRs close
* Post a PR summary comment with deployment/job/image details

It also adds deployment setup documentation in `deployments/CI_DEPLOYMENTS.md` and updates `env.example` to reflect the current local/frontend configuration.

Note This does NOT handle frontend build or deploy, which will be handled in a separate PR

PR #70, opened 2026-04-23
Frontend deploy

AlexAxthelm:
Add deployment process for frontend building on #63, and #68.

Move node-build to CI pipeline, and includes step in deploy process to write `config.json` with actual values for API/other services

### DIFF
--- a/.github/workflows/azure-static-web-apps.yml
+++ b/.github/workflows/azure-static-web-apps.yml
@@ -120,9 +120,6 @@ jobs:
           test -f .swa-dist/config.json
           test -f .swa-dist/staticwebapp.config.json
 
-      - name: Install OIDC Client from Core Package
-        run: npm install @actions/core@1.6.0 @actions/http-client
-
       - name: Deploy
         id: builddeploy
         uses: Azure/static-web-apps-deploy@v1
--- a/.github/workflows/azure-static-web-apps.yml
+++ b/.github/workflows/azure-static-web-apps.yml
@@ -1,0 +1,176 @@
+name: Azure Static Web Apps CI/CD
+
+on:
+  workflow_call:
+    inputs:
+      artifact-name:
+        description: "Uploaded frontend artifact name"
+        required: true
+        type: string
+      app-env:
+        description: "Runtime app environment label"
+        required: true
+        type: string
+      api-url:
+        description: "Base API URL to inject into runtime config"
+        required: true
+        type: string
+      entity-linkage-url:
+        description: "Base entity linkage URL to inject into runtime config"
+        required: true
+        type: string
+      auth0-domain:
+        description: "Auth0 domain"
+        required: true
+        type: string
+      auth0-client-id:
+        description: "Auth0 client ID"
+        required: true
+        type: string
+      auth0-audience:
+        description: "Auth0 audience"
+        required: true
+        type: string
+
+jobs:
+  build_and_deploy_job:
+    runs-on: ubuntu-latest
+    name: Build and Deploy Job
+    permissions:
+      checks: write
+      contents: read
+      id-token: write
+      pull-requests: write
+    steps:
+
+      - uses: actions/checkout@v5
+        with:
+          submodules: true
+          lfs: false
+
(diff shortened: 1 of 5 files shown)


.github/workflows/node-build.yml
```diff
@@ -41,3 +39,25 @@ jobs:
...
+        run: |
+          if [[ -f dist.tgz ]]; then
+            echo "has_artifact=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_artifact=false" >> "$GITHUB_OUTPUT"
+          fi
+
+        # Reuse the artifact for downstream deploy
+      - name: Upload dist.tgz as workflow artifact
+        if: ${{ steps.check_artifact.outputs.has_artifact == 'true' }}
```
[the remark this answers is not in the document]
AlexAxthelm: want to reserve this functionality, since dependabot PRs won't run this sucessfully (lesson from PBTAR)

AlexAxthelm:
@jdhoffa Noting that the Required "Build / Build Frontend" check has been moved to be "CD Pipeline / Build frontend / Build Frontend"

.github/workflows/azure-static-web-apps.yml
```diff
@@ -0,0 +1,173 @@
...
+        uses: Azure/static-web-apps-deploy@v1
+        with:
+          azure_static_web_apps_api_token: ${{ secrets.AZURE_STATIC_WEB_APPS_DEPLOY_TOKEN }}
+          action: "upload"
+          skip_app_build: true
+          app_location: ".swa-dist"
+          skip_api_build: true
+
+      - name: Write comment JSON
+        if: ${{ github.event_name == 'pull_request' }}
```
jdhoffa, line 134: This new SWA deploy job is a reusable workflow (`on: workflow_call`), but it still checks `github.event_name == 'pull_request'` inside the called workflow. In a reusable workflow, I believe the triggering event is `workflow_call`, so these PR-only steps may never run?

.github/workflows/azure-static-web-apps.yml
```diff
@@ -0,0 +1,173 @@
...
+                "service": $service,
+                "url": $url
+              }
+            }' > "$json_file"
+
+          jq empty "$json_file"
+          cat "$json_file"
+
+      - name: Upload comment JSON artifact
+        if: ${{ github.event_name == 'pull_request' }}
```
jdhoffa, line 168: Same here

.github/workflows/node.yml
```diff
@@ -24,7 +24,3 @@ jobs:
   format:
     name: "Formatting"
     uses: ./.github/workflows/node-format.yml
-
-  build:
-    name: "Build"
-    uses: ./.github/workflows/node-build.yml
```
jdhoffa, line 30: NB: This PR removes the standalone frontend build from Node Package Checks. The frontend still builds later in CD Pipeline, but that couples basic build validation to deployment-oriented CI instead of the normal Node checks.

If branch protection only keys off the Node workflow, a frontend build break could now slip past the usual fast-feedback path.

We may consider keeping it here, understanding that it's just a build check?

jdhoffa (changes requested):
Potentially blocking comments. One non-blocking

.github/workflows/azure-static-web-apps.yml
AlexAxthelm: That's correct. We also call it on `push` to main, for example. The PR-only step is commenting on the PR, so it only makes sense to include in that context, since for `workflow_call` or `push`, there isn't a thread to write to.

.github/workflows/node.yml
AlexAxthelm: Sure. I believe this technically runs the same step twice, but I don't have a particular problem with that here.

.github/workflows/azure-static-web-apps.yml
AlexAxthelm: In this case, it's because we haven't set up a "PROD" environment for stitch yet (auth + real data)

jdhoffa (approved):

PR #72, opened 2026-04-27
Don't run seed job if no new DB

AlexAxthelm:
Adds output to DB deploy process to signal if new database was created. If not, skip the `seed` job.

Resolves issue in PR preview environments where `seed` would run regularly, and overpopulate the merge review (since all resources have matches)

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -146,9 +146,10 @@ jobs:
 
   run-seed:
     name: "Run seed"
+    if: ${{ needs.deploy-db.outputs.database-created == 'true' }}
     uses: ./.github/workflows/run-seed.yml
     secrets: inherit
-    needs: [build_seed_docker_image, deploy-api]
+    needs: [build_seed_docker_image, deploy-db, deploy-api]
     with:
       full-image-name: ${{ needs.build_seed_docker_image.outputs.digest-image-name }}
       api-url: ${{ format('{0}/api/v1', needs.deploy-api.outputs.container-app-url) }}
--- a/.github/workflows/deploy-db.yml
+++ b/.github/workflows/deploy-db.yml
@@ -2,6 +2,9 @@
 on:
   workflow_call:
     outputs:
+      database-created:
+        description: "Whether this workflow created the database during this run"
+        value: ${{ jobs.create-db.outputs.database-created }}
       postgres-host:
         description: "Database server hostname"
         value: ${{ jobs.create-db.outputs.postgres-host }}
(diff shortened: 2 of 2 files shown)


AlexAxthelm:
from deployed API:
```
curl -X 'GET' \
  'https://pr72-api.purplegrass-c07d0a94.westus2.azurecontainerapps.io/api/v1/oil-gas-fields/?sort_by=name&sort_order=asc&page=1&page_size=5' \
  -H 'accept: application/json'
```

```
...
  "total_count": 34,
  "page": 1,
  "page_size": 5,
  "total_pages": 7
```

AlexAxthelm:
Same response from API (34 items), and showing skipped job in GH UI:
Screenshot_2026-04-27_14 39 36@2x

.github/workflows/build-and-deploy.yml
```diff
@@ -146,9 +146,10 @@ jobs:
 
   run-seed:
     name: "Run seed"
+    if: ${{ needs.deploy-db.outputs.database-created == 'true' }}
```
jdhoffa, line 149: (Review supported by `codex`):
Skipping run-seed based only on “database was created in this run” breaks recovery after any partial/failed seed run. This new gate only allows seeding when `deploy-db.yml` reports a fresh CREATE DATABASE. But the seed job is not resumable: it POSTs payloads one by one and aborts on the first non-2xx. If the first PR run creates the DB and then seed fails midway, every later rerun/push sees `database-created=false` and permanently skips seed, leaving that preview DB empty or partially populated.

A possible fix is to gate `run-seed` on a seed-specific marker, not on overall DB creation.
Currently “DB exists” is being used as a proxy for “seed completed successfully”, and those are not equivalent. A failed or interrupted seed run leaves the DB existing but unseeded.

So for example:
  1. Make `run-seed` idempotent or at least self-detecting.
  2. Have it check for a sentinel that means “seed already applied successfully”.
  3. Only skip when that sentinel exists.

jdhoffa: @AlexAxthelm up to you if you want to treat this as blocking or non-blocking. but I thought it was relevant to flag.

AlexAxthelm: @jdhoffa I'm going to treat this as non-blocking, mostly because it's a behavior that I've already noted, but I don't want to put effort into solving it, since we have already flagged a more robust backup/restore/migration/DBA stuff pass as something we'll need to handle before we get to PROD. CC @mbarlow12

jdhoffa (approved):

PR #73, opened 2026-04-29
Llm retrieval (STIT-441)

AlexAxthelm:
Adds a `stitch-llm` service which can GET resource details from the API, and forward, along with a structured prompt, to OpenAI Foundry, which returns a structured JSON response.

Key differences from #60:
* LLM responses must have a citation available for any value they are suggesting, for human review
* LLM stands as a separate service, giving us a layer of access control around how data (non-public) is handled while being passed to the LLM, and continuing to use the standard access model
* LLM responses include the original prompt and response from the Foundry

Screenshot_2026-04-29_14 18 24@2x


AlexAxthelm:
Noting that this is still in draft, largely because I'm seeing a challenge in getting the structured output (JSON) and the citations in the same response. The OpenAI instance seems to attach citations for its web search (tool use) to _unstructured_ (free text) responses, but I don't see a great way to handle that right now for something that needs to pass validation. cc @jdhoffa

jdhoffa:
> Adds a `stitch-llm` service which can GET resource details from the API, and forward, along with a structured prompt, to OpenAI Foundry, which returns a structured JSON response.
> 
> Key differences from #60:
> 
> * LLM responses must have a citation available for any value they are suggesting, for human review
> * LLM stands as a separate service, giving us a layer of access control around how data (non-public) is handled while being passed to the LLM, and continuing to use the standard access model
> * LLM responses include the original prompt and response from the Foundry
> 
> Screenshot_2026-04-29_14 18 24@2x

WOOHOOOOOO

AlexAxthelm:
Screenshot_2026-05-05_10 57 27@2x

deployments/stitch-llm/src/stitch/llm/routers/oil_gas_fields.py
```diff
@@ -0,0 +1,150 @@
...
+    return FieldSuggestionResponse(
+        resource_id=id,
+        field=field,
+        value=value,
+        citations=citations,
+        query_succeeded=True,
+        model=llm_result.model,
+        rationale=parsed.rationale,
+        foundry_request=llm_result.request_payload,
+        foundry_response=llm_result.response_payload,
```
jdhoffa, line 149: Probably NB, but worth discussing:
Right now the API exposes ‘foundry_request’ and ‘foundry_response’ in the browser. 

while I totally 100% agree that we should be tracking and logging that info, I’m not sure we want it to be available to any end user?

What do you think?

Request will contain raw source data. All that source data should be public so that’s probably fine

Response data may contain info about the token usage and whatnot… not sure if there’s anything that would be truly sensitive in the response

jdhoffa (approved):

AlexAxthelm:
Merged, noting that the response model has not been finalized yet. In particular, as we further consider how to persist this to DB, we will probably need to expose less (the full LLM response/query) in the user-level response to the query.

PR #74, opened 2026-04-29
feat(api): add source-filtered resource lists (STIT-448)

mbarlow12:
Adds source-aware resource list queries with priority coalescing, redaction-aware provenance, and owners/operators in list responses.

Also tightens query-param validation and adds coverage for filtering, sorting, counts, redacted sources, and JSON list fall-through.

### Redaction + filtering
Primary logic is implemented as a series of CTEs that exhibit a few different layers:
1. Base/`qualified` CTE: filters source rows from query & ensures redacted sources rows are NULL-filled
2. flat/`coalesced` CTE: groups by resource id, uses `case` statement for priority handling
    - Note that at this point, "redacted" values are NULL
    - Special handling applied to JSON columns
3. Builds query conditions and applies to `coalesced` CTE
4. Extracts totals and applies pagination logic
5. constructs & the list of entities and total

- `source` from query params limits which active memberships participate
  - i.e. "get all resources connected to a `gem` source"  or "get resources connected to a `gem` or `rmi` source"
- Redaction is applied after `source` filtering
- `OUTER JOIN` on sources allows us to keep a resource in the result set/count, but their field values and provenance are treated as missing/NULL
- Final text/exact filters, sorting, counts, and pagination run against the coalesced visible values.
- Null or redacted higher-priority sources fall through to lower-priority visible sources.
- Empty `owners`/`operators` lists still count as real values.

Tested with full API pytest, ruff, format check, and basedpyright.


deployments/api/src/stitch/api/entities.py
jdhoffa: `sort_by="source"` is still part of the shared `SortableField` schema in, but the resource endpoint now rejects it at runtime in `deployments/api/src/stitch/api/db/og_field_resource_actions.py`. That means the OpenAPI contract for `/oil-gas-fields/` still advertises source as valid even though requests get a `422`. This should probably be split into separate sort schemas for resources vs sources.

## Repro steps
1. Open https://pr74-api.purplegrass-c07d0a94.westus2.azurecontainerapps.io/docs
2. Expand `GET /oil-gas-fields/`.
3. Click Try it out.
4. In `sort_by`, enter `source`.
5. Execute.
6. Observe `422` Error

Image

jdhoffa (dismissed):
See one blocking comment

deployments/api/src/stitch/api/entities.py
mbarlow12: Fixed. "source" removed from "sort_by"

jdhoffa (approved):
Looks good!

deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
```diff
@@ -139,7 +140,7 @@ def _base_query[QM: OGFieldQueryMixin](
         stmt: Select[tuple[QM]] = select(cls).distinct()
         for cond in cls._build_conditions(params):
             stmt = stmt.where(cond)
-        return cls._apply_sort(stmt, params)
+        return stmt.order_by(*cls._create_sort_clauses(params))
```
AlexAxthelm, line 143: NB: For my understanding, this is still in query construction?

AlexAxthelm (approved):
Looks good from my side

deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
mbarlow12: Correct. We can probably merge the mixin back into the original model since its not being used by a view any longer.

PR #76, opened 2026-05-04
Update .gitignore to ignore .codex directory

jdhoffa:
Add .codex directory to .gitignore to exclude local context files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -220,3 +220,7 @@ __marimo__/
 
 # macOS
 **/.DS_Store
+
+# codex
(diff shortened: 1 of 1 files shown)


AlexAxthelm (approved):
No technical blocker here, and it might make sense for use to have this for tools that we officially support and are commonly used by the team, but as a style/preference, I would say that developer-specific tooling exclusions are probably placed in either `/git/info/exclude`, or more likely, a developer's global `.gitignore` (so they exclude it system-wide, rather than in a particular repo)

jdhoffa:
Hmmm global gitignore sounds like a good alternative

PR #78, opened 2026-05-06
feat(auth): extract permissions claim from access tokens [STIT-334]

mbarlow12:
## Summary

- `TokenClaims` now exposes a `permissions: frozenset[str]` field
- `JWTValidator` reads Auth0's `permissions` array claim (rfc9068_profile_authz dialect)
- Missing/null/empty → empty frozenset; malformed shape → `TokenValidationError`
- Extract-only — no enforcement here. Wiring `permissions` into `stitch-api` route guards is a follow-up.

## Test plan

- [x] `pytest packages/stitch-auth` — 40 passed (14 new)
- [x] `make api-test-exact` — 134 passed, no consumer regression
- [x] `make check` — lint/format/type/frontend all green
- [ ] smoke against a real Auth0 token in the running app

### DIFF
--- a/packages/stitch-auth/src/stitch/auth/validator.py
+++ b/packages/stitch-auth/src/stitch/auth/validator.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+from typing import Any
 
 import jwt
 from jwt import PyJWKClient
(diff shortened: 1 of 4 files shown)


AlexAxthelm (approved):
Looks good. I opened #79 to allow us to inspect the permissions and other token info in the UI.

packages/stitch-auth/src/stitch/auth/validator.py
```diff
@@ -47,10 +68,12 @@ def validate(self, token: str) -> TokenClaims:
 
         email = payload.get("email") or payload.get("preferred_username")
```
AlexAxthelm, line 69: I'm testing with my actual bearer token (from our current Auth0 setup) and noting that the token doesn't have `name` or `email`, which has been part of why we've been leaning so much on AUTH_DISABLED to this point. Just-in-time provisioning cannot work with the tokens in their current form

AlexAxthelm (changes requested):
Making this part blocking until we sort out the JIT provisioning issue with Auth0 tokens. CC @jdhoffa

packages/stitch-auth/src/stitch/auth/validator.py
AlexAxthelm: Specifically the error that's causing a problem here is:
```
api-1             |   File "/app/.venv/lib/python3.12/site-packages/psycopg/cursor_async.py", line 117, in execute
api-1             |     raise ex.with_traceback(None)
api-1             | psycopg.errors.NotNullViolation: null value in column "name" of relation "users" violates not-null constraint
api-1             | DETAIL:  Failing row contains (6, auth0|698dd5fe4fd7f2f3d1a27db4, null, null).
````

which I'm getting on when I "fetch" resources through the UI where I've authenticated. Wondering if only requiring username/email for write operations might be a better call?

AlexAxthelm (approved):
Talked to @mbarlow12 about handling the missing values in the JWT, and the option we're going to pursue is dropping the name/email requirement for JIT provisioning. We'll cover that in a separate PR, which unblocks this (since CI envs are AUTH_DISABLED for now)

PR #80, opened 2026-05-07
feat: user name & email nullable

mbarlow12:
- Makes `name` and `email` nullable on `users` table. User identity is now pegged just to `sub` and `id`, but since no auditing features exist (yet), we can handle this later.
- provisions users in separate session/transaction so users still persist if the rest of the request fails

**Verification**
 - additional unit tests
- Full local setup with `AUTH_DISABLED=false` & hack to add my personal bearer token to the `seed` container's env (not committed). Seed populates, entity linkage an LLM backfill fully functional.

### DIFF
--- a/deployments/api/src/stitch/api/entities.py
+++ b/deployments/api/src/stitch/api/entities.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from enum import StrEnum
 from typing import Literal
 
-from pydantic import BaseModel, EmailStr, Field, computed_field
+from pydantic import BaseModel, Field, computed_field
 
 from stitch.ogsi.model import GEM_SRC, LLM_SRC, RMI_SRC, WM_SRC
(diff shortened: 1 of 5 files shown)


AlexAxthelm (approved):
Confirmed that running with `AUTH_DISABLED=false` allows my actual Auth0 token to JIT provision a user on the `users` table

deployments/api/src/stitch/api/entities.py
```diff
@@ -34,8 +34,8 @@ class User(BaseModel):
     id: int = Field(...)
     sub: str = Field(...)
     role: str | None = None
-    email: EmailStr
-    name: str
+    email: str | None = None
```
AlexAxthelm, line 37: NB: Noting that the `EmailStr` validation is probably still what we want here. Added back in #85

PR #81, opened 2026-05-07
feat(api): gate source data on licensed permission claim [STIT-334]

mbarlow12:
Wires the `resource:read:licensed:{src}` Auth0 claim through the API. Inverts the dormant `redacted_sources` blacklist into a `licensed_sources` allowlist, applied to resource list, resource detail, and source endpoints. Resources with no licensed memberships still appear, just as null shells.

Out of scope (tracked separately): `preview_merge_candidate`, write-path permissions, frontend "no licensed source" affordance.

### DIFF
--- a/deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
+++ b/deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from collections.abc import Sequence
+from collections.abc import Collection, Sequence
 from typing import Any, ClassVar, Self
 
 from sqlalchemy import (
@@ -24,6 +24,7 @@ from stitch.api.entities import OGFieldQueryParams, OGSI_SOURCE_DEFAULT
 from stitch.ogsi.model import LocationType
 from stitch.ogsi.model.types import (
     FieldStatus,
+    OGSISrcKey,
     PrimaryHydrocarbonGroup,
     ProductionConventionality,
 )
@@ -103,9 +104,10 @@ class OGFieldQueryMixin:
         cls,
         session: AsyncSession,
         params: OGFieldQueryParams,
+        licensed_sources: Collection[OGSISrcKey] | None = None,
     ) -> Sequence[Self]:
         """Execute a filtered, sorted, paginated query and return (rows, total)."""
-        base = cls._base_query(params)
+        base = cls._base_query(params, licensed_sources=licensed_sources)
         stmt = cls._apply_pagination(base, params)
         rows = (await session.scalars(stmt)).all()
         return rows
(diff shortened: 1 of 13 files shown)


deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
```diff
@@ -115,12 +117,15 @@ async def count(
         cls,
         session: AsyncSession,
         params: OGFieldQueryParams | None = None,
+        licensed_sources: Collection[OGSISrcKey] | None = None,
```
AlexAxthelm, line 120: NB: Shouldn't the count be the same for licensed or unlicensed? the difference would be in if it's NULL or not? or is this taking filters into account?

deployments/api/src/stitch/api/permissions/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    """Return the set of OGSI sources the caller is licensed to read."""
+    out: set[OGSISrcKey] = set()
+    for perm in claims.permissions:
+        if not perm.startswith(_LICENSED_PREFIX):
+            continue
+        candidate = perm[len(_LICENSED_PREFIX) :]
+        if candidate in _VALID_SOURCES:
+            out.add(cast(OGSISrcKey, candidate))
+        else:
+            logger.debug("ignoring unknown source in permission: %r", perm)
```
AlexAxthelm: NB: Should this be a WARN?

AlexAxthelm (approved):
Looks good. Tested with a new Auth0 Account with reduced permissions (GEM+RMI, right side in screenshot) 

Image

deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
mbarlow12: The intended behavior is that the query params determine the returned rows, but they apply *after* redaction & coalescing. So for no filters/query, the counts should be the same. But a licensed user should get different counts if say "state = Texas" came from `wm`.

This is also making me realize that the `filter-options` endpoint should probably use the same query logic instead of the underlying view.

I spent some time yesterday trying to simplify the query...haven't really landed on a decent solution. The combination of:
- need to apply queries the flattened resource row
- need to extract provenance per attribute/column
- disallowed source rows need to be null *before* flattening
- **per column** `CASE`, `COALESCE`, or window functions to do the flattening

make for a **huge** SQL statement (really, really big). I'm not expert enough in Postgres to know if that's necessarily a problem, but it smells a bit to me.

I think if the query is functioning as intended, we can leave it for now but keep an eye on it. Maybe we'll have another "use the Bearer token in the env" moment and it'll be solved :rofl:.

deployments/api/src/stitch/api/permissions/__init__.py
mbarlow12: good catch. fixed

deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
AlexAxthelm: Yeah, there's a smell there, and I think the "DBA" answer would be more normalizing forms, but that brings its own costs. What we have now is pretty close to where I would have landed with it, so my vote is keep it as is (easy to reason about) and let the PG query optimizer get a workout

AlexAxthelm (approved):

PR #86, opened 2026-05-08
feat: add M2M auth flow to `stitch-client` and ensure appropriate API handling

mbarlow12:
> [!NOTE]
> Merge #80 before this PR.

Adds a minimal auth token flow to the `stitch-client` package. Package users/consumers (seed, entity_linkage, llm) can be given credentials for `client_id` (Auth0 "Application") and `client_secret`, allowing them to make a call to the RMI Auth0 tenant. Posting to `/oauth/token` will return the `access_token` that can be used to make authenticated calls to our API.

- `stitch-api` treats all callers as "users" without distinguishing between a machine/service and a human user
- `stitch-client` can still be used with user pass-through credentials so a service like `entity_linkage` can operate on a user's behalf


mbarlow12:
Closing. Going to reassess with more careful design.

PR #88, opened 2026-05-11
Refresh resource list and source data views (ui refresh 1)

jdhoffa:
**This PR was supported by `codex`.**

This PR refreshes the Resources view with automatic loading, a Refresh action, resource count/filter/sort summaries, and diagnostics for endpoint/page/sort details. 

Improves the resource table rendering with normalized field access, sortable `field_status`, and cleaner filter/dropdown styling.
Replaces the raw JSON resource output with structured rendering, including a Source data section with source records.
Clarifies the provenance indicators with visible field-level source labels, accessible source mix summaries, and unknown-source handling.

I also remove the stand-alone single-resource fetch demo from the home route so the page focuses on the Resource list.


deployments/stitch-frontend/src/components/StructuredDataView.jsx
```diff
@@ -0,0 +1,240 @@
...
+          <dd className="mt-1 break-words text-sm text-ink">
+            <ValueText value={value} />
+          </dd>
+        </div>
+      ))}
+    </dl>
+  );
+}
+
+function getPrimitiveColumns(rows) {
```
jasonbrownrmi, line 98: StructuredDataView silently drops columns past 
  **MAX_TABLE_COLUMNS (8) — StructuredDataView.jsx:80-90**

  If a row has 12 primitive fields, only the first 8 are shown and the rest disappear — no "Details" indicator, no fallback. 

  The fallback check at StructuredDataView.jsx:176-181 already does this:

  if (primitiveColumnCount > 0 && primitiveColumnCount <=
  MAX_TABLE_COLUMNS) {
    return <DataTable rows={data} />;
  }

getPrimitiveColumns caps at 8 and returns early, so primitiveColumnCount can never exceed 8 — meaning the fallback branch never triggers for the "too many columns" case. Either remove the early return so the count reflects reality, or have it return a sentinel when the cap is exceeded.

deployments/stitch-frontend/src/components/StructuredDataView.jsx
```diff
@@ -0,0 +1,240 @@
...
+}
+
+function DataTable({ rows }) {
+  const columns = getPrimitiveColumns(rows);
+
+  if (!columns.length) {
+    return (
+      <div className="space-y-3">
+        {rows.map((row, index) => (
+          <StructuredDataInner key={index} data={row} depth={1} />
```
jasonbrownrmi: Redundant work in DataTable -  StructuredDataView.jsx:104-138

The "any row has nested fields?" check runs twice per render (once in <thead>, once per row in <tbody>), and each row also re-filters Object.entries(row) to build the Details cell. Compute once:

  const detailColumns = rows.map(row =>
    Object.entries(row).filter(([k, v]) => !columns.includes(k) &&
  !isPrimitive(v))
  );
  const hasDetails = detailColumns.some(d => d.length > 0);

Worth doing when this component renders large mock arrays.

deployments/stitch-frontend/src/components/ResourcesView.jsx
```diff
@@ -15,107 +14,181 @@ import {
...
+      sort_by: sortConfig.column ?? undefined,
+      sort_order: sortConfig.column ? sortConfig.direction : undefined,
+    },
+  );
 
-  const handleFetch = () => {
-    setEnabled(true);
-    refetch();
-  };
+  const resources = useMemo(() => data?.items ?? [], [data]);
```
jasonbrownrmi: data?.items is already a stable reference for a given data object from useQuery, and useMemo with [data] only re-evaluates when data changes. The wrapping adds nothing. Same comment in useResources callers throughout. Just const resources = data?.items ?? []

deployments/stitch-frontend/src/components/ResourcesView.jsx
```diff
@@ -15,107 +14,181 @@ import {
...
+    (count, values) => count + values.length,
+    0,
+  );
 
-  const handleClear = () => {
-    queryClient.removeQueries({ queryKey: resourceKeys.lists(endpoint) });
-    setEnabled(false);
-    setFilters(EMPTY_FILTERS);
-    setSearchParams({});
+  const handleRefresh = () => {
```
jasonbrownrmi, line 60: handleRefresh double-invokes the fetch

  queryClient.invalidateQueries({ queryKey:
  resourceKeys.lists(endpoint) });
  refetch();

  invalidateQueries already triggers a refetch on active observers - calling refetch() after it usually causes a second network request. Pick one (probably just refetch() since the query is already active).

deployments/stitch-frontend/src/components/StructuredDataView.jsx
```diff
@@ -0,0 +1,240 @@
+import { FIELD_META } from "../constants/fieldMeta";
+
+const MAX_TABLE_COLUMNS = 8;
+
+function isPrimitive(value) {
```
jasonbrownrmi: isPrimitive defined twice - StructuredDataView.jsx and resourceDisplay.js Move it into resourceDisplay.js (or a utils/typeGuards.js) and import.

deployments/stitch-frontend/src/hooks/useResources.js
```diff
@@ -79,14 +84,84 @@ function useMergeCandidateReal(
...
+    }),
+  );
+}
+
+function compareMockResources(a, b, field, sortOrder = "asc") {
+  const aValue = getResourceField(a, field);
+  const bValue = getResourceField(b, field);
+  const direction = sortOrder === "desc" ? -1 : 1;
+
+  if (aValue == null && bValue == null) return a.id - b.id;
```
jasonbrownrmi: Sort stability uses numeric subtraction on potentially-non-numeric IDs — useResources.js

  if (aValue == null && bValue == null) return a.id - b.id;

If id is ever a string (UUID, API value), this returns NaN. The mock data may always be numeric, but the assumption is implicit. Either coerce (Number(a.id) - Number(b.id)) or do a string compare.

deployments/stitch-frontend/src/utils/resourceDisplay.js
```diff
@@ -0,0 +1,53 @@
...
+  return resource?.data?.[key] ?? resource?.[key] ?? null;
+}
+
+function isPrimitive(value) {
+  return (
+    value == null || ["string", "number", "boolean"].includes(typeof value)
+  );
+}
+
+function deriveProvenance(resource) {
```
jasonbrownrmi: deriveProvenance is O(fields × sources × records) and first-source-wins.

For each primitive field, it scans every source's records to find a match. For mock data this is fine; if it ever gets exercised on real API payloads it could be slow. Maybe worth a comment that the SOURCES iteration order determines the winner when the same value appears in multiple sources - it's a non-obvious behavior baked into ordering.

deployments/stitch-frontend/src/utils/resourceDisplay.js
```diff
@@ -0,0 +1,53 @@
...
+  return provenance;
+}
+
+export function normalizeResourceListItem(resource) {
+  if (!resource || resource.data) {
+    return resource;
+  }
+
+  const data = { ...resource };
+  delete data.id;
```
jasonbrownrmi: deletes a hard-coded blacklist:

  delete data.id;
  delete data.source_data;
  delete data.repointed_to;
  delete data.constituents;
  delete data.provenance;

Consider a whitelist of fields that belong in data instead, or extract the blacklist as a named constant so the intent is documented.

deployments/stitch-frontend/src/components/ResourcesList.jsx
```diff
@@ -1,43 +1,67 @@
+import { useMemo, useState } from "react";
```
jasonbrownrmi: ResourcesList.jsx does not get imported anyware but is still being modified in the PR.

deployments/stitch-frontend/src/components/StructuredDataView.jsx
```diff
@@ -0,0 +1,240 @@
...
+              Object.entries(row).some(
+                ([key, value]) => !columns.includes(key) && !isPrimitive(value),
+              ),
+            ) ? (
+              <th className="px-3 py-2">Details</th>
+            ) : null}
+          </tr>
+        </thead>
+        <tbody className="divide-y divide-line">
+          {rows.map((row, rowIndex) => (
```
jasonbrownrmi: StructuredDataView uses array indices as React keys - StructuredDataView.jsx:130, 188

  {rows.map((row, rowIndex) => <tr key={rowIndex}>…

Fine for static data, but if the same component is fed re-ordered or partially-updated arrays it'll re-mount rows unnecessarily. If rows have an id, use that instead; otherwise leave a comment that the data is rendered as a static snapshot.

deployments/stitch-frontend/src/components/StructuredDataView.jsx
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    return (
+      <div className="space-y-3">
+        {data.map((item, index) => (
+          <section
+            key={index}
+            className="min-w-0 border-t border-line pt-3 first:border-t-0 first:pt-0"
+          >
+            <h4 className="mb-2 text-sm font-semibold text-ink">
```
jasonbrownrmi: Small deal, but <h4> heading tag here could create heading-level skips if this component is nested without h1, h2, h3 headers above. Might consider a heading level prop.

deployments/stitch-frontend/src/components/StructuredDataView.jsx
```diff
@@ -0,0 +1,240 @@
+import { FIELD_META } from "../constants/fieldMeta";
```
jasonbrownrmi, line 1: No tests for this component. Consider adding some.

deployments/stitch-frontend/src/utils/resourceDisplay.js
```diff
@@ -0,0 +1,53 @@
+import { SOURCES } from "../constants/sourceMeta";
```
jasonbrownrmi, line 1: No tests for normalizeResourceListItem and deriveProvenance for this utility. Consider adding some.

jasonbrownrmi (commented):
Hi Jackson, here are some comments on the PR. Hope this helps.

deployments/stitch-frontend/src/components/StructuredDataView.jsx
jdhoffa: Closed by 40f9094

deployments/stitch-frontend/src/utils/resourceDisplay.js
jdhoffa: Closed by 40f9094

deployments/stitch-frontend/src/components/StructuredDataView.jsx
jdhoffa: Resolved by 790067f

deployments/stitch-frontend/src/components/StructuredDataView.jsx
jdhoffa: Resolved by 790067f

deployments/stitch-frontend/src/components/StructuredDataView.jsx
jdhoffa: Resolved by 790067f

deployments/stitch-frontend/src/components/ResourcesView.jsx
jdhoffa: Resolved by 790067f

deployments/stitch-frontend/src/components/ResourcesView.jsx
jdhoffa: Resolved by 790067f

deployments/stitch-frontend/src/components/StructuredDataView.jsx
jdhoffa: Resolved by 790067f

deployments/stitch-frontend/src/components/StructuredDataView.jsx
jdhoffa: Resolved by 790067f

deployments/stitch-frontend/src/components/ResourcesList.jsx
jdhoffa: Resolved by 790067f

jasonbrownrmi (approved):

PR #89, opened 2026-05-11
Refresh merge candidate review workflow (ui refresh 2)

jdhoffa:
*Depends on #88*
*This PR was supported by `codex`*

This PR reworks the Merge Review with a focused candidate decision panel, including the addition of status badges, review/pending/total counts, merge preview rendering, and decision notes. 

It also updates Entity Linkage to use the new Structured Data display mechanism (introduced in #88).


deployments/stitch-frontend/src/pages/EntityLinkagePage.jsx
```diff
@@ -39,73 +42,72 @@ export default function EntityLinkagePage() {
...
       ) : null}
 
       <section>
-        <h2 className="mb-2 text-lg font-medium">Response JSON</h2>
-        <pre className="overflow-x-auto rounded border bg-gray-50 p-4 text-sm">
-          {JSON.stringify(responseJson, null, 2)}
-        </pre>
+        <h2 className="mb-2 text-lg font-semibold text-ink">Run result</h2>
+        <div className="rounded-md border border-line bg-panel p-4">
+          <StructuredDataView
```
AlexAxthelm: Image 

We may want some better visuals handlings here? it's hard to tell where one group ends and the next begins.

AlexAxthelm (commented):
Looks nice!

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
jasonbrownrmi: Redundant useEffect() - Tanstack Query already refetches when:
  - useMergeCandidates(ENDPOINT, true) mounts with enabled: true
  - useMergeCandidate(ENDPOINT, selectedId, …)'s queryKey changes (because selectedId is in the key)
  - useMergeCandidatePreview(...)'s enabled flag flips

This useEffect fires a second refetch on top of Tanstack Query's own. Can just be deleted - the hooks already do the right thing.

jasonbrownrmi: Redundant useEffect() - Tanstack Query already refetches when:
  - useMergeCandidates(ENDPOINT, true) mounts with enabled: true
  - useMergeCandidate(ENDPOINT, selectedId, …)'s queryKey changes (because selectedId is in the key)
  - useMergeCandidatePreview(...)'s enabled flag flips

This useEffect fires a second refetch on top of Tanstack Query's own. Can just be deleted - the hooks already do the right thing.

jasonbrownrmi: Redundant useEffect() - Tanstack Query already refetches when:
  - useMergeCandidates(ENDPOINT, true) mounts with enabled: true
  - useMergeCandidate(ENDPOINT, selectedId, …)'s queryKey changes (because selectedId is in the key)
  - useMergeCandidatePreview(...)'s enabled flag flips

This useEffect fires a second refetch on top of Tanstack Query's own. Can just be deleted - the hooks already do the right thing.

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
```diff
@@ -139,196 +507,85 @@ export default function MergeCandidateReviewPage() {
           queryKey: resourceKeys.mergeCandidate(ENDPOINT, candidate.id),
         }),
         queryClient.invalidateQueries({
-          queryKey: resourceKeys.mergeCandidatePreview(ENDPOINT, candidate.id),
+          queryKey: resourceKeys.preview(ENDPOINT, candidate.id),
         }),
       ]);
 
       refetchCandidates();
       refetchCandidate();
```
jasonbrownrmi: This function invalidates and explicitly refetches. invalidateQueries already triggers refetches on active observers. The trailing two refetch() calls duplicate the network requests. Drop the explicit refetch()s (the highlighted block).

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
```diff
@@ -114,11 +471,22 @@ export default function MergeCandidateReviewPage() {
...
+      .length ?? 0;
+
+  function handleSelect(id) {
+    setSelectedId(id);
+    setReviewNotes("");
+    setActionError(null);
+    setShowSourceResources(false);
+  }
 
   async function handleReview(action) {
```
jasonbrownrmi, line 468: No auto-advance after a successful review.

After approve/deny:
- selectedId stays the same
- The candidate's status flips PENDING → APPROVED/DENIED, so DecisionControls disappears
- The user is left staring at a now-reviewed card with no next action

The auto-select effect at 431-436 (!selectedId && candidates?.length) only runs once at mount. If the intent is "review one at a time," advancing to the next PENDING after a successful review feels like the natural finish - otherwise the user has to click back to the queue every time. 

Consider:
```
const nextPending = candidates?.find(c => c.id !== candidate.id && c.status === "PENDING");
if (nextPending) setSelectedId(nextPending.id);
```
Either way, calling setReviewNotes("") after success would also clean up; right now stale notes persist on the same candidate (not user-visible because controls hide, but cleared selection state is safer).

deployments/stitch-frontend/src/pages/EntityLinkagePage.jsx
```diff
@@ -39,73 +42,72 @@ export default function EntityLinkagePage() {
...
+      setError({
+        message: err instanceof Error ? err.message : String(err),
+      });
     } finally {
       setLoading(false);
     }
   }
 
   return (
-    <main className="mx-auto max-w-5xl p-6">
```
jasonbrownrmi, line 64: Drops the <main> landmark — EntityLinkagePage.jsx

If the app shell (e.g., in App.jsx or a layout component) wraps routes in <main>, this is fine. Otherwise it's an accessibility regression: screen-reader users lose the main-content landmark for this page. MergeCandidateReviewPage is also a plain <div>; same question applies.

deployments/stitch-frontend/src/pages/EntityLinkagePage.jsx
```diff
@@ -39,73 +42,72 @@ export default function EntityLinkagePage() {
...
       }
 
-      setResponseJson(parsed);
+      setResult(parsed);
     } catch (err) {
-      setError(err instanceof Error ? err.message : String(err));
-      setResponseJson(null);
+      setError({
+        message: err instanceof Error ? err.message : String(err),
+      });
```
jasonbrownrmi, line 146: Inconsistent error shape.

  // path A: HTTP non-OK
  setError({ status: response.status, body: parsed });
  // path B: caught exception
  setError({ message: err instanceof Error ? err.message : String(err) });

Two different shapes flowing into the same renderer. Works because StructuredDataView is shape-agnostic, but consumers reading the error UI see "status: 401, body: …" sometimes and "message: …" other times. Pick one.

```
catch (err) {
  setError({ status: null, body: err instanceof Error ? err.message : String(err) });
}
```

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
```diff
@@ -14,34 +12,391 @@ import { createAuthenticatedFetcher } from "../auth/api";
...
+  actionError,
+  actionLoading,
+  activeReviewAction,
+  shouldShowPreview,
+  preview,
+  previewLoading,
+  previewError,
+  previewErrorObj,
+  showSourceResources,
+  onToggleSourceResources,
```
jasonbrownrmi, line 296: Not blocking but this is a lot of props and worth thinking about in the future. 

One option is to pass candidateQuery and previewQuery as objects ({ data, isLoading, isError, error }) instead of flattening each, which would reduce prop count by 8.

jasonbrownrmi (approved):
Just a few minor comments for this one.

deployments/stitch-frontend/src/pages/EntityLinkagePage.jsx
jdhoffa: Resolved by 605747b

deployments/stitch-frontend/src/pages/EntityLinkagePage.jsx
jdhoffa: Resolved by 962e211

deployments/stitch-frontend/src/pages/EntityLinkagePage.jsx
jdhoffa: Resolved by 962e211

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
jdhoffa: Resolved by 962e211

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
jdhoffa: Resolved by 962e211

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
jdhoffa: Resolved by 962e211

jdhoffa: Resolved by 962e211

jdhoffa: Resolved by 962e211

deployments/stitch-frontend/src/pages/MergeCandidateReviewPage.jsx
jdhoffa: Resolved by 0359c39

PR #90, opened 2026-05-12
Add resource browser search to UI (ui refresh 3) [STIT-95]

jdhoffa:
*Depends on #89*
*This PR was supported by `codex`*

Adds resource browser search controls to the refreshed Resources UI. Wires submitted search text through `useResources`, query keys, and API query params as `q`.

Supersedes #77


AlexAxthelm (approved):
Looks good from my side. Noting that this is passing the `q` param to the api, and making a new request to list resources

PR #98, opened 2026-05-13
chore: update dependencies & remove `click` usage (STIT-451)

mbarlow12:
Fixes dependabot alerts on `urllib3` and `PyJWT`.
- the `standard-no-fastapi-cloud-cli` optional dependency removes `urllib3`

The only  `click` usages were in `__main__.py` modules that were never called, so they're dropped along with any `[project.scripts]` that used them.


deployments/api/pyproject.toml
```diff
@@ -6,7 +6,7 @@ readme = "README.md"
 authors = [{ name = "Michael Barlow", email = "mbarlow@rmi.org" }]
 requires-python = ">=3.12.12"
 dependencies = [
-  "fastapi[standard]>=0.124.0",
+  "fastapi[standard-no-fastapi-cloud-cli]>=0.135.1",
```
AlexAxthelm, line 9: NB: Does `uv` have anything like rust's `foo.workspace = true` to keep the various subpackages aligned?

AlexAxthelm (approved):

deployments/api/pyproject.toml
mbarlow12: Aligned how, exactly?

It's what we're already doing at the top-level `pyproject.toml` with `[tool.uv.workspace]` and `[tool.uv.sources]`. We declare the paths where `uv` should look for member packages and then whether, for a given package name, `uv` should try to resolve it in the workspace or some other index.

As far as I know, there's no clean way to declare global/shared dependencies. We hack around it a bit right now with `pytest` and `ruff` at the repo root (though we're actually declaring those individually on each member package). Admittedly, the intended design of workspaces is to have a single target (usually under `src/`) and the member packages serve as libraries. It adds friction but also reinforces clearer isolation and separation of concerns when the target becomes sufficiently complex.

`stitch` breaks this to some degree right now with several separate deployments that are also member packages. I do think we could get to a place were a "deployment" is effectively a pyproject.toml and a Dockerfile, and we compose the necessary member packages to get the functionality we want.

mbarlow12: E.g. we could declare "entity linkage" logic in the `stitch-ogsi` package. It doesn't talk to a db or request/response shapes and only deals in its own models/types. Then, either `stitch-api` or `stitch-entity-linkage` adds it as a dependency, handles adapting/transforming the data into correct shapes and calls on the `stitch-ogsi` entity linkage interface to actually perform the business logic of deciding which resources should be merged.

PR #109, opened 2026-05-28
feat: add permissions strings and checks (STIT-422 & STIT-423)

mbarlow12:
Adds the updated permissions strings to `stitch-auth` since this is shared between all our deployments.

Provides `check_permissions` which can be used/called within a deployment.

### DIFF
--- a/packages/stitch-auth/src/stitch/auth/permissions.py
+++ b/packages/stitch-auth/src/stitch/auth/permissions.py
@@ -3,8 +3,10 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Collection, Iterable
-from typing import TypeAlias
+from collections.abc import Callable, Collection, Iterable
+from typing import Literal, NoReturn, TypeAlias
+
+from .errors import InsufficientPermissionsError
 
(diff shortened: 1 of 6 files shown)


packages/stitch-auth/src/stitch/auth/permissions.py
```diff
@@ -0,0 +1,162 @@
...
+SOURCE_READ_LLM: Permission = f"{SOURCE_READ_PREFIX}llm"
+SOURCE_READ_PERMISSIONS: frozenset[Permission] = frozenset(
+    {
+        SOURCE_READ_RMI,
+        SOURCE_READ_GEM,
+        SOURCE_READ_WM,
+        SOURCE_READ_LLM,
+    }
+)
+SOURCE_WRITE: Permission = "source:write"
```
AlexAxthelm, line 31: Do we want to restrict `source:write` to source types? (`source:write:gem`)

AlexAxthelm (approved):
No major issue from my side, but copilot flagged a few

AlexAxthelm:
@mbarlow12 Merging as-is, with open notes above.

PR #110, opened 2026-05-28
feat: integrate auth permissions (STIT-422 & STIT-423)

mbarlow12:
Small permissions wiring pass for STIT-422/423.

- adds shared stitch-auth permission strings + guard
- wires API routes through require_permissions(...), including source-read check="any"
- keeps source licensing backed by the shared parser


AlexAxthelm (approved):
LGTM

PR #119, opened 2026-06-03
Feat/alembic migrations (STIT-427)

AlexAxthelm:
* Add alembic dependency
* Wire up alembic to handle migrations instead of current init job ("baseline")
* Remove outdated `init_job.py`
* Update membership model to use FK constraint, to avoid having sources without a matching priority identified
* Use alembic migration in CI/CD pipeline
* Add "autogenerate" target in makefile


deployments/api/pyproject.toml
```diff
@@ -6,6 +6,7 @@ readme = "README.md"
 authors = [{ name = "Michael Barlow", email = "mbarlow@rmi.org" }]
 requires-python = ">=3.12.12"
 dependencies = [
+  "alembic>=1.17.1",
```
mbarlow12, line 9: NB: If we add `alembic` as a optional dependency (i.e. `uv add alembic --optional migrations`), then we can add a smidgen of environment isolation for the migrator container. We could either build a second image or run `uv sync --extra migrations` before running `alembic` stuff.

deployments/api/src/stitch/api/alembic.py
```diff
@@ -0,0 +1,179 @@
...
+        command.upgrade(config, settings.revision)
+
+    _run_with_connection(
+        operation_name=f"upgrade to {revision}",
+        revision=revision,
+        runner=_upgrade,
+    )
+
+
+def run_autogenerate(message: str = "baseline") -> None:
```
mbarlow12: I don't see a whole lot of benefit in adding the alembic features here over running `uv run --package stitch-api alembic revision`. I actually think minimizing how much is automated here will help us more tightly manage migrations.

.github/workflows/run-alembic.yml
```diff
@@ -74,21 +74,21 @@ jobs:
             -e POSTGRES_PASSWORD \
             -e PGSSLMODE \
             "${{ inputs.full-image-name }}" \
-            python -m stitch.api.db.init_job
+            python -m stitch.api.alembic_upgrade head
```
mbarlow12: Curious on the advantage of this over calling `alembic upgrade head` or `uv run alembic upgrade head`.?

deployments/api/src/stitch/api/db/model/membership.py
```diff
@@ -28,7 +28,11 @@ class MembershipModel(TimestampMixin, UserAuditMixin, Base):
     resource_id: Mapped[int] = mapped_column(
         ForeignKey("og_field_resources.id"), nullable=False
     )
-    source: Mapped[OGSISrcKey] = mapped_column(String(10), nullable=False)
+    source: Mapped[OGSISrcKey] = mapped_column(
+        String(10),
+        ForeignKey("og_field_source_priority.source"),
+        nullable=False,
+    )
```
mbarlow12, line 35: NB: At one point, I could've sworn this was a `Literal` type that gets translated into an `enum` on the PG side.

So we have it recorded somewhere, this makes me wonder if we might want a `source_metadata` table instead of only priority: `id, key, default_priority, ...other stuff?, required_permission?` (dunno exactly). If we're going to move on attribute selection in the relatively near future, this might be an adjustment that could help us.

deployments/api/src/stitch/api/alembic.py
```diff
@@ -0,0 +1,179 @@
...
+def _env(name: str, default: str = "") -> str:
+    value = os.environ.get(name)
+    return default if value is None else value
+
+
+def build_db_url() -> str:
+    url = os.environ.get("DATABASE_URL")
+    if url:
+        return url
+    return ApiSettings().get_sync_database_url().render_as_string(hide_password=False)
```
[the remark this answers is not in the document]
mbarlow12: So this is 100% not true, but it does need to be configured. https://alembic.sqlalchemy.org/en/latest/cookbook.html#using-asyncio-with-alembic

We might see migrations run a bit faster since async would allow multiple network calls to be in-flight, but I don't know enough about how the PG server handles that scenario. Alembic has transaction isolation, so if PG can queue up multiple transactions, then I'd more strongly advocate for this.

deployments/api/src/stitch/api/alembic.py
```diff
@@ -0,0 +1,179 @@
...
+            logger.info("alembic %s complete.", operation_name)
+        finally:
+            if settings.use_advisory_lock:
+                logger.info("releasing advisory lock...")
+                release_lock(conn)
+
+    engine.dispose()
+
+
+def run_upgrade(revision: str = "head") -> None:
```
mbarlow12: Similar to https://github.com/RMI/stitch/pull/119/changes#r3360033775. `alembic/env.py` is intentionally flexible to allow us to customize `run_migrations_online/run_migrations_offline`. I think it makes sense to pull more classes/functions we've already written in `deployments/api/src` and configure the connection more directly in `env.py`.

mbarlow12 (commented):
Some misc comments, but the overall gist is that I'd advocate for leaning more heavily on the `alembic` CLI, definitely for creating revision files and in our workflows. We'd add more imports into `alembic/env.py` (from `stitch.api....`), but we'd have much more explicit separation between our application code and managing migrations.

I think it'd also give us more flexibility & freedom to put out fires or quickly patch DB errors either directly from our local machine (not ideal), setting arguments in workflows, or other fanciness  in our CI/CD. We'd get some additional hygiene in instituting that all migrating happens from `[uv run] alembic upgrade ...`.

deployments/api/tests/test_alembic_revisions.py
```diff
@@ -0,0 +1,45 @@
...
+    return False
+
+
+def test_all_alembic_revisions_have_explicit_irreversible_downgrade():
+    revision_files = sorted(REVISION_DIR.glob("*.py"))
+
+    assert revision_files, f"No Alembic revision files found in {REVISION_DIR}"
+
+    for revision_file in revision_files:
+        tree = ast.parse(revision_file.read_text(), filename=str(revision_file))
```
mbarlow12, line 31: NB: you could also use `importlib` to import the module and call the function.

mbarlow12 (approved):
👍

PR #124, opened 2026-06-09
Add alembic check to CI

AlexAxthelm:
Adds 2 CI checks:

* CD Pipeline / Check DB Migrations (clean)
* CD Pipeline / Check DB Migrations (live)

These run the alembic check operation (after migrations are applied), to ensure consistency between the migration files and what actually gets applied to DBs (i.e. ensure that migration files don't end in an inconsistent/malformed state). 

`(clean)` runs a throwaway postgres instance (in docker) and migrates and checks against that.

`(live)` waits for migrations to run against the deployed DB (PR-124 DB here) and ensures that checks correctly (check against vendor-specific issues)


jdhoffa (approved):
Love it!

PR #135, opened 2026-06-17
Feat/db optimizations (STIT-502)

AlexAxthelm:
Major refactor to DB, so that rather than storing data values for sources in "wide" format in `oil_gas_field_sources`, it's stored "long" in `oil_gas_field_source_values`, allowing for more idiomatic SQL queries.

Python data model untouched, but there are some changes to the SQLAlchemy translation layer between the two data models.


deployments/api/src/stitch/api/db/model/oil_gas_field_source_value.py
```diff
@@ -0,0 +1,163 @@
...
+assert set(ATTRIBUTE_KINDS) == set(OilGasFieldBase.model_fields), (
+    "ATTRIBUTE_KINDS out of sync with OilGasFieldBase fields: "
+    f"{set(ATTRIBUTE_KINDS) ^ set(OilGasFieldBase.model_fields)}"
+)
+
+ATTRIBUTE_NAMES: tuple[str, ...] = tuple(ATTRIBUTE_KINDS)
+
+# Attributes stored as JSON (owners/operators) -- emitted as text by the list
+# coalescing pivot and deserialized in Python.
+JSON_ATTRIBUTE_NAMES: frozenset[str] = frozenset(
```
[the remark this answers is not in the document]
AlexAxthelm: this is used in other files

deployments/api/src/stitch/api/db/coalesce_sql.py
```diff
@@ -0,0 +1,222 @@
...
+            r.id.label("resource_id"),
+            m.source.label("source"),
+            m.source_pk.label("source_pk"),
+            priority.label("priority"),
+        )
+        .select_from(m)
+        .join(r, r.id == m.resource_id)
+        .join(p, p.source == m.source)
+        .outerjoin(o, and_(o.resource_id == r.id, o.source == m.source))
+        .where(
```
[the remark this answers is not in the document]
AlexAxthelm: https://github.com/RMI/stitch/pull/135/commits/ea47e94dbd221d3e56da1250f353ab8e541777d8

deployments/api/src/stitch/api/db/coalesce_sql.py
```diff
@@ -0,0 +1,222 @@
...
+            m.source.in_(list(dict.fromkeys(selected_sources)))
+        )
+    if licensed_sources is not None:
+        active_src = active_src.where(
+            m.source.in_(list(dict.fromkeys(licensed_sources)))
+        )
+    if resource_ids is not None:
+        active_src = active_src.where(r.id.in_(list(resource_ids)))
+    # `s` join keeps the header reachable and confirms the source row exists.
+    active_src = active_src.join(s, s.id == m.source_pk).cte("active_src")
```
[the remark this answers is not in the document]
AlexAxthelm: https://github.com/RMI/stitch/pull/135/commits/ea47e94dbd221d3e56da1250f353ab8e541777d8

deployments/api/alembic/versions/f3fb36006ce6_baseline.py
```diff
@@ -309,6 +264,83 @@ def upgrade() -> None:
...
+        sa.Column(
+            "value_json",
+            sa.JSON(none_as_null=True).with_variant(
+                postgresql.JSONB(none_as_null=True, astext_type=sa.Text()), "postgresql"
+            ),
+            nullable=True,
+        ),
+        sa.CheckConstraint(
+            "colname IN ('name', 'country', 'name_local', 'state_province', 'region', 'basin', 'reservoir_formation', 'location_type', 'production_conventionality', 'primary_hydrocarbon_group', 'field_status', 'latitude', 'longitude', 'discovery_year', 'production_start_year', 'fid_year', 'owners', 'operators')",
+            name="ck_source_value_colname",
```
mbarlow12, line 320: Good constraints to add, but could present some difficulties/gotchas if the model shifts. Mainly requires refreshing/recreating the constraint in subsequent migrations–looks like it could be easy to miss.

deployments/api/alembic/versions/f3fb36006ce6_baseline.py
```diff
@@ -309,6 +264,83 @@ def upgrade() -> None:
         ),
         sa.PrimaryKeyConstraint("id"),
     )
+    op.create_table(
+        "og_field_resource_source_priority",
```
mbarlow12, line 268: If we add a `source_id` column here (or replace `source` w/ `source_id`), then we can embed the default priorities in code, and when a resource is created/updated we then update the priority table.

This would mean that all top level resources would have at least 1 row in the `og_field_resource_source_priority` table, and we could drop the static source priority table. It adds some complexity but also sets the groundwork for user-mediated priorities.

I guess it seems like if we're going to undertake the effort to allow for priority overrides, getting more granular at the source id level gets us more flexibility to pick specific source rows where multiple source keys are present.

deployments/api/src/stitch/api/db/model/og_field_query_mixin.py
```diff
@@ -119,96 +144,63 @@ async def count(
...
 
-        source_col = getattr(cls, "source", None)
-        if source_col is not None:
-            sources = list(
-                dict.fromkeys(getattr(params, "source", OGSI_SOURCE_DEFAULT))
-            )
-            conditions.append(source_col.in_(sources))
-            if licensed_sources is not None:
-                conditions.append(source_col.in_(list(dict.fromkeys(licensed_sources))))
+        sources = list(dict.fromkeys(getattr(params, "source", OGSI_SOURCE_DEFAULT)))
```
mbarlow12: FWIW, we did discuss dropping this from the query params. I think it makes sense as "filter resources that have some X source attached" doesn't seem all that valuable.

deployments/api/src/stitch/api/db/model/og_field_resource_source_priority.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+    # FUTURE: key the override on a specific source record (FK to
+    # oil_gas_field_sources.id) rather than the source *key*, so a resource with
+    # multiple records from the same source (e.g. two WoodMac records) can be
+    # ranked individually. Today priority is per source-class.
+    source: Mapped[OGSISrcKey] = mapped_column(
+        String(10),
+        ForeignKey("og_field_source_priority.source"),
+        primary_key=True,
+    )
```
mbarlow12, line 33: See comment above, but I'd vote to go ahead and bite the bullet on `source_id` and drop the default table entirely.

The future work I'd see is adding a `source_value_id` or another structure (`source_id`, `colname`) to point directly to the EAV table/view rows, allowing users to set priorities on individual attributes/columns.

deployments/api/src/stitch/api/db/coalesce_sql.py
```diff
@@ -0,0 +1,234 @@
...
+    spine = _resource_spine(selected_sources).cte("resource_spine")
+    coalesced = select(spine.c.id.label("id"))
+    for field_name in ATTRIBUTE_NAMES:
+        coalesced = coalesced.add_columns(
+            pivot.c[field_name],
+            pivot.c[f"{field_name}{PROVENANCE_SUFFIX}"],
+        )
+    coalesced = coalesced.select_from(
+        spine.outerjoin(pivot, pivot.c.resource_id == spine.c.id)
+    )
```
mbarlow12, line 196: The "spine" element here looks redundant. Since we're not doing any filtering at this point, it seem like the joins and where clauses (incl `selected_sources`) are already captured in the `values_cte`.

Maybe I'm missing something, but why couldn't we return `pivot.cte("licensed_resource_list")`?

deployments/api/src/stitch/api/db/coalesce_sql.py
```diff
@@ -0,0 +1,234 @@
...
+        value = materialize_value(
+            colname,
+            value_text=row["value_text"],
+            value_num=row["value_num"],
+            value_json=row["value_json"],
+        )
+        view_data[colname] = value
+        provenance[colname] = (value, row["source"], row["source_pk"])
+
+    return OilGasFieldBase(**view_data), provenance
```
mbarlow12, line 234: If `resource_id` is a required field, it seems like building the entire CTE is pretty heavy-handed vs a more typical select/get. Couldn't we essentially reuse the same logic as `og_field_resource_actions.py#get`?

deployments/api/alembic/versions/f3fb36006ce6_baseline.py
AlexAxthelm: Yeah. we discussed this, but I dropped it in implementation. I think it's a pretty easy change when we implement the rest of the workflow, and is a good use case to try an actual migration.

deployments/api/src/stitch/api/db/coalesce_sql.py
AlexAxthelm: I think that you might be right, but I'm inclined to flag as something to come back to?

mbarlow12 (approved):

PR #136, opened 2026-06-18
Add check for PLAN files, to warn against merging

AlexAxthelm:
Adds check (non-required) that there are no PLAN files in the repo. A normal idiom would be to add to `.gitignore`, but it's useful to commit the PLAN at the beginning of a branch, as a record of the intent, and then clean after (especially on multi-step plans). This permits that workflow, while also flagging the cleanup need as part of the PR process.


.github/workflows/admin-no-plan-file.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Ensure no plan files are tracked
+        shell: bash
+        run: |
+          # Plan files are useful to commit alongside a change set as a record
+          # of intent, but must be removed before merging. This fails if any
+          # matching file is tracked by git on this branch.
+          #
+          # Matches (case-insensitive), at the repo root or any subdirectory:
+          #   PLAN.md, PLAN.<anything>.md  (e.g. PLAN.feature-x.md)
+          found="$(git ls-files | grep -iE '(^|/)PLAN(\.[^/]+)*\.md$' || true)"
```
jdhoffa: **NB:** `|| true` here swallows both grep's expected exit-1-on-no-match *and* any failure from `git ls-files`. With `shell: bash` GHA enables `pipefail`, so if git ever breaks in this step, the check silently passes. Could split into two steps for a tiny robustness win:

```bash
tracked="$(git ls-files)"
found="$(printf '%s\n' "$tracked" | grep -iE '(^|/)PLAN(\.[^/]+)*\.md$' || true)"
```

Not worth gating on.

.github/workflows/admin-no-plan-file.yml
```diff
@@ -0,0 +1,33 @@
...
+
+          if [ -n "$found" ]; then
+            echo "::error::Found committed plan file(s). Remove them before merging:"
```
jdhoffa: **NB:** Both this line and the per-file `::error file=$f::` on line 28 emit annotations, so each offender shows up twice in the PR Files tab — once as a generic workflow error, once file-scoped. The per-file annotation alone is probably enough; consider demoting this summary to a plain `echo` (or `::notice::`). Pure polish.

jdhoffa (commented):
A few non-blocking comments.

Drive-by (not for this PR): `admin.yml:18` has a preexisting typo in the concurrency group — `github.event_action` should be `github.event.action`. Worth a follow-up.

.github/workflows/admin-no-plan-file.yml
AlexAxthelm: https://github.com/RMI/stitch/pull/136/commits/95667e54d94b10a86e3485f4b3142a5271ff819c

.github/workflows/admin-no-plan-file.yml
AlexAxthelm: https://github.com/RMI/stitch/pull/136/commits/95667e54d94b10a86e3485f4b3142a5271ff819c

jdhoffa (approved):

PR #134, opened 2026-06-17
Feat/observability

AlexAxthelm:
Adds an observability layer, and ties into middleware and DB queries, to capture performance information about both.

Deliberately simple at this point, but generally compatible with transitioning to OTel later.

See `deployments/PERFORMANCE.md` for more detailed instructions on capture and analysis, but highlights:
* Emits JSON structured records
* Can filter samples in API (only emit for long queries), configurable to capture all
* Can make requests `X-Perf-Scenario: <label>` request header to tag requests and their related queries for grouping
* Includes basic analysis script

Most code by claude.

PR #141, opened 2026-06-18
feat: query optimization spike

mbarlow12:
For the spike, we've added a new table (for sqlite compatibility) with a minimal set of sync actions to keep data up-to-date with resources, sources, and memberships. In PG, this would make more sense as a materialized view with incremental updates on write. Using a view would also remove the need for the additional sync functionality.

There's also a stress test script at `deployments/api/scripts/test_query.py` that profiles the query over 6 test cases. It builds (and optionally persists) a sqlite db file, seeds it with random, reasonably representative data, and runs the queries.

Over ~80k sources & resources with multiple layers of merging, I'm seeing 0.3-0.9s for all 6 variants.

Overall table/view structure is essentially identical to #135 for the main query table. It's used exclusively to fetch the appropriate set of resource ids that can then be used to hydrate the actual response.

Note: this is a large changeset since it's purely additive. We can integrate by swapping `query` in `og_field_resource_actions.py` with the `v2` call before fully deleting the old implementation and tests. Majority of added lines come from `scripts/test_query.py` and additional modules under `tests/`.

---

Implemented mostly by Claude Code.


deployments/api/scripts/test_query.py
AlexAxthelm, line 1: There's a lot to like about this test script. Loading data directly into sqlite for testing is fast (but does couple the test script a bit tighter to the DB schema). I think a lot of this could get integrated in followups to #134

deployments/api/src/stitch/api/db/model/og_field_resource_query_view.py
```diff
@@ -0,0 +1,70 @@
...
+from sqlalchemy import ForeignKey, Index, Integer, String
+from sqlalchemy.orm import Mapped, mapped_column
+from stitch.ogsi.model.og_field import OilGasFieldBase
+
+from .common import Base
+from .types import PORTABLE_BIGINT, PORTABLE_JSON, PORTABLE_REAL
+
+
+# Field → value-column routing constants (single source of truth)
+VALUE_TEXT_FIELDS = (
```
AlexAxthelm, line 12: You landed on largely the same boilerplate EAV construction I did in #135. Being boring is good on this.

deployments/api/src/stitch/api/db/og_field_query_view_refresh.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if rows:
+        await session.execute(insert(OGFieldResourceQueryView), rows)
+
+    await session.flush()
+
+
+async def rebuild_all(session: AsyncSession, *, batch_size: int = 1000) -> None:
+    """Full rebuild of the projection (used by stress script and backfill)."""
+    await session.execute(delete(OGFieldResourceQueryView))
```
AlexAxthelm, line 104: I'm not totally sure how SQLAlchemy would handle this, but throwing around a `delete` on a table like this can cause a lot of problems on a Postgres instance (where `CASCADE` exists). More idiomatic would be to delete all rows (but keep the table), or as you pointed out, MV.

AlexAxthelm (changes requested):
I don't have any particular changes to request in code, but I think it's worth discussing the architectural differences from my parallel PR around the same issue (#135)

In either case, we're introducing and Entity-Attribute-Value table, which allows for fast query, since it's the SQL idiom. 

The major difference I see is that you're keeping the primary store in the current `source` table, and then computing what is effectively a search/filter index in the `OGFieldResourceQueryView`, where I'm using the EAV as a primary data store. 

Of the two, I have a preference for my design, since it avoids the need to maintain any syncing utilities (which are workarounds for SQLite not having Materialized Views?) at a modest read-time cost, and introducing a break between the storage model and logical model (I'm not entirely certain what the performance trade off is, I haven't run any formal benchmarking yet).

The other big difference is that you're keeping the current baseline migration, while I'm willing to start over in the face of a big restructuring

Marking as "changes requested" mostly to raise discussion on how we want to go here (cc @jdhoffa )

PR #142, opened 2026-06-22
Ci/OpenTelemetry

AlexAxthelm:
Adds OpenTelemetry (auto)instrumentation to API and adds related deployments.

Note that at this point, OTel is OPT-IN: include it in the stack through the new `make reboot-docker-heavy` target, which works like `reboot-docker`, but alsoo stands up an OTel collecter to capture data, and forward to the Jaeger all-in-one which enables analysis, and can be accessed in browser at `http://localhost:16686/`

Also note that this has not been configured for our cloud deployments yet (local only), but that will be coming in a soon PR


jdhoffa (commented):
Reviewed the OpenTelemetry wiring end-to-end. One blocking issue (env-var prefix mismatch in `env.example`) plus a handful of non-blocking nits flagged inline with `NB:`. The overall shape is good — clean module separation, sensible defaults, sampling done right, and the heavy stack opt-in is well isolated.

jdhoffa (commented):
(stale, see below)

jdhoffa (commented):
Reviewed the OpenTelemetry wiring end-to-end. One blocking issue (env-var prefix mismatch in `env.example`) plus a handful of non-blocking nits flagged inline with `NB:`. The overall shape is good — clean module separation, sensible defaults, sampling done right, and the heavy stack opt-in is well isolated.

env.example
```diff
@@ -7,6 +7,27 @@ STITCH_LLM_LOG_LEVEL=info
...
+#   otlp              - spans are sent to the collector -> Jaeger (UI on
+#                       http://localhost:16686). `make reboot-docker-heavy`
+#                       brings up the collector + Jaeger AND flips the API to
+#                       otlp for you, so you do NOT need to set this to otlp here
+#                       for the heavy stack. Set it manually only when pointing a
+#                       host `make api-dev` at an already-running heavy stack
+#                       (and then also set API_OTEL_OTLP_ENDPOINT to localhost).
+#   none              - tracing disabled.
+# OTEL_SAMPLE_RATIO: root-trace sampling ratio (1.0 = capture everything).
+API_OTEL_ENABLED=true
```
jdhoffa, line 22: **Blocking.** `Settings` (`deployments/api/src/stitch/api/settings.py:91-95`) has no `env_prefix`, so the `API_OTEL_*` names declared here are not read by pydantic when running on the host (e.g. `make api-dev`). Inside Docker they work only because `docker-compose.yml` rewrites them to `OTEL_*` before injecting into the container. The example on line 25 (`# API_OTEL_OTLP_ENDPOINT=http://localhost:4317`) is specifically the host workflow, and as written it silently does nothing.

Two ways to fix:
- Give `Settings` (or a nested `OtelConfig`) `env_prefix="API_"` so the prefixed names work everywhere, **or**
- Rename these to the unprefixed forms (`OTEL_ENABLED`, `OTEL_TRACES_EXPORTER`, `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_SAMPLE_RATIO`) — which also matches what the OTel SDK reads natively.

Whichever path you pick, the override hint in `deployments/api/conftest.py:9` should line up with it.

deployments/api/conftest.py
```diff
@@ -0,0 +1,14 @@
+"""Suite-wide environment setup, applied before any app import.
+
+Loaded by pytest from the rootdir before ``tests/conftest.py`` (which imports
+the app, and the app configures OpenTelemetry tracing at import time).
+
+Tracing is process-global and emits span records to stdout on every request;
+the unit/integration suite doesn't exercise it, and benchmarking/trace
+verification runs against real infrastructure instead. Disable it here. Override
+by exporting ``OTEL_TRACES_EXPORTER=console`` when debugging tracing locally.
```
jdhoffa, line 9: NB: This docstring tells the reader to export `OTEL_TRACES_EXPORTER=console`, which is correct *for current `Settings` behavior* (no prefix) but inconsistent with `env.example`, which uses `API_OTEL_*`. Tied to the env-prefix blocker on `env.example:22` — fix the prefix story once and update both.

deployments/api/src/stitch/api/main.py
```diff
@@ -31,16 +31,23 @@ async def lifespan(app: FastAPI):
     app.state.auth_config_validated = True
     yield
     await dispose_engine()
+    if _tracer_provider is not None:
```
jdhoffa, line 34: NB: `_tracer_provider` is read here but assigned ~8 lines below (line 42). Python resolves the name at call time so this works in practice, but if module import raises between the `lifespan` def and the assignment, shutdown will `NameError`. Cheap fix: declare `_tracer_provider: TracerProvider | None = None` above the `lifespan` definition, then reassign after `configure_tracing`.

deployments/api/src/stitch/api/observability/tracing.py
```diff
@@ -0,0 +1,133 @@
...
+    sampler = ParentBased(root=TraceIdRatioBased(settings.otel_sample_ratio))
+    provider = TracerProvider(resource=resource, sampler=sampler)
+
+    if settings.otel_traces_exporter == "otlp":
+        # endpoint=None lets the exporter fall back to OTEL_EXPORTER_OTLP_ENDPOINT
+        # / the localhost default.
+        exporter = OTLPSpanExporter(endpoint=settings.otel_exporter_otlp_endpoint)
+        provider.add_span_processor(BatchSpanProcessor(exporter))
+    else:  # "console" — log spans to stdout, no sidecar required.
+        provider.add_span_processor(SimpleSpanProcessor(LoggingSpanExporter()))
```
jdhoffa, line 117: NB: `SimpleSpanProcessor` exports synchronously per span on the request thread. With FastAPI + SQLAlchemy auto-instrumentation, a request that issues 20 queries now does 20 synchronous JSON serializations through `logging.info` (plus the server span) before responding. Fine for local dev, but worth measuring before this default is used anywhere that cares about latency. `BatchSpanProcessor` here too (with the existing `provider.shutdown()` on lifespan exit doing the flush) would remove that hot-path cost.

deployments/api/src/stitch/api/observability/tracing.py
```diff
@@ -0,0 +1,133 @@
...
+                        "span_name": span.name,
+                        "trace_id": format(ctx.trace_id, "032x"),
+                        "span_id": format(ctx.span_id, "016x"),
+                        "parent_span_id": format(parent.span_id, "016x")
+                        if parent is not None
+                        else None,
+                        "kind": span.kind.name,
+                        "duration_ms": duration_ms,
+                        "status": span.status.status_code.name,
+                        "attributes": dict(span.attributes or {}),
```
jdhoffa, line 82: NB: `dict(span.attributes or {})` copies the full attribute bag onto the log record. SQLAlchemy auto-instrumentation puts the rendered SQL in `db.statement`, which on bulk inserts or large `IN (...)` queries can be very large. On the console path these end up in stdout JSON records and then in Log Analytics. Consider an allowlist of attribute keys, or a max-length truncation pass before logging.

deployments/api/src/stitch/api/observability/tracing.py
```diff
@@ -0,0 +1,133 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from fastapi import FastAPI
+    from opentelemetry.sdk.trace import ReadableSpan
+    from sqlalchemy.engine import Engine
+
+    from ..settings import Settings
+
+SERVICE_NAME = "stitch-api"
```
jdhoffa, line 45: NB: `SERVICE_NAME` is hardcoded. If you ever run a second service that imports this module (or want to override per-deployment without a code edit), driving this from `settings.service_name` (default `"stitch-api"`) keeps the door open.

deployments/api/src/stitch/api/db/config.py
```diff
@@ -58,6 +58,9 @@ def get_engine() -> AsyncEngine:
         slow_query_ms=settings.slow_query_ms,
         log_all_queries=settings.log_all_queries,
     )
+    # Per-query spans (separate from the aggregate timing listener above).
+    if settings.otel_enabled and settings.otel_traces_exporter != "none":
```
jdhoffa, line 62: NB: Worth a one-line comment that `get_engine` is `@lru_cache`'d, so this instrumentation runs once per process (and re-runs cleanly after `dispose_engine()` clears the cache and a new engine is built). Took me a minute to convince myself there's no double-instrumentation hazard.

deployments/api/tests/observability/test_tracing.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class TestConfigureTracing:
+    def test_returns_none_when_disabled(self):
+        assert configure_tracing(Settings(otel_enabled=False)) is None
+
+    def test_returns_none_when_exporter_is_none(self):
+        assert configure_tracing(Settings(otel_traces_exporter="none")) is None
+
+    def test_console_exporter_returns_provider(self):
```
jdhoffa, line 69: NB: `configure_tracing` calls `trace.set_tracer_provider(...)`, which mutates global OTel state. The `try/finally` here shuts the provider down but doesn't unset the global, so the next test (in this file or elsewhere in the suite) that calls `configure_tracing` will get OTel's `Overriding of current TracerProvider is not allowed` warning. The conftest-level `OTEL_TRACES_EXPORTER=none` masks this in production paths, but the test itself is the thing exercising the global mutation.

Either construct `TracerProvider(...)` directly here without going through `configure_tracing`, or add a fixture that swaps `trace._TRACER_PROVIDER` back in `finally`.

deployments/api/tests/observability/test_tracing.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_emits_one_record_per_span_with_core_fields(self, logging_tracer, caplog):
+        with caplog.at_level(logging.INFO, logger=_TRACE_LOGGER):
+            with logging_tracer.start_as_current_span("my-span") as span:
+                span.set_attribute("stitch.request_id", "abc123")
+
+        events = _trace_events(caplog)
+        assert len(events) == 1
+        event = events[0]
+        assert event["span_name"] == "my-span"
+        assert event["kind"] == "INTERNAL"
```
jdhoffa, line 43: NB: Asserting against the string literals `"INTERNAL"` (here) and `"UNSET"` (line 48) means an OTel SDK rename would produce an opaque failure. Pulling these from `SpanKind.INTERNAL.name` / `StatusCode.UNSET.name` is more resilient and reads as "this should match what the SDK produces" rather than "this exact string".

docker-compose.otel.yml
```diff
@@ -0,0 +1,29 @@
...
+# so the normal dev path never pulls these images. Layered on via
+# `make reboot-docker-heavy`.
+#
+# To send traces here, set the API's exporter to OTLP:
+#   API_OTEL_TRACES_EXPORTER=otlp
+# Then open the Jaeger UI at http://localhost:16686.
+services:
+
+  otel-collector:
+    image: otel/opentelemetry-collector-contrib:latest
```
jdhoffa, line 12: NB: `:latest` for the collector means a `make reboot-docker-heavy` after an upstream release can change collector behavior with no other change in the repo. Pinning to a major or a specific minor (e.g. `:0.110.0` or `:0`) trades some drift for reproducibility. Same for `jaegertracing/all-in-one:latest` on line 24.

docker-compose.otel.yml
```diff
@@ -0,0 +1,29 @@
...
+      - ./deployments/otel-collector/config.yaml:/etc/otel-collector-config.yaml:ro
+    ports:
+      - "4317:4317"   # OTLP gRPC
+      - "4318:4318"   # OTLP HTTP
+    profiles: ["full"]
+    depends_on:
+      - jaeger
+
+  jaeger:
+    image: jaegertracing/all-in-one:latest
```
jdhoffa, line 24: NB: See the comment on line 12 — same `:latest` concern applies here. Jaeger UI / behavior can change between releases (e.g. they shifted to v2 of the binary recently).

docker-compose.otel.yml
AlexAxthelm: In theory, yes, but jaeger seems fairly stable, and the collecter is something that we probably want to keep up to date, since this is the image that includes that Azure wiring too. (Re-evaluate in a few weeks?)

deployments/api/src/stitch/api/observability/tracing.py
AlexAxthelm: Fair, and as this gets extracted, it probably will get parameterized more. in particular `stitch-service` in #143 also exposes a similar variable

env.example
```diff
@@ -7,6 +7,27 @@ STITCH_LLM_LOG_LEVEL=info
...
+API_OTEL_TRACES_EXPORTER=console
+# For host `make api-dev` against a running heavy stack, point at the mapped port:
+# API_OTEL_OTLP_ENDPOINT=http://localhost:4317
+API_OTEL_OTLP_ENDPOINT=http://otel-collector:4317
+API_OTEL_SAMPLE_RATIO=1.0
```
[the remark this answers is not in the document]
AlexAxthelm: Addressed

env.example
AlexAxthelm: Addressed with safer defaults in https://github.com/RMI/stitch/pull/142/commits/d3900bd1f166ea0daf3a52062e42115672be0111

jdhoffa (approved):

mbarlow12 (commented):
Just hitting a post-merge comment here.

It seems like this might be a good candidate for a shared package. We can use optional/extra dependencies where the observability stack needs `sqlalchemy`,  `fastapi`, etc...

If we're planning to use the same set of tracing functions, this could help with code reuse and with consistency across deployments.

PR #144, opened 2026-06-24
STIT-506 - add core geography and field attributes

jasonbrownrmi:
## Summary

Adds Country, Field status, and Primary hydrocarbon group to the Resource List as sortable columns and filter dropdowns, and displays country as its conventional name (e.g. NOR → "Norway") instead of the raw ISO alpha-3 code. The supporting API already accepted these as sort/filter/filter-option fields, so this is a frontend-only change.

### Acceptance criteria

  - [x] Resource List can be filtered by
  country, field_status, and
  primary_hydrocarbon_group
  - [x] All three are sortable and filterable
  - [x] Existing filtering/sorting behavior is
  unchanged (additive only)

#### Columns & filters
  - ResourcesTable.jsx - adds Country and Primary hydrocarbon group columns (both sortable); introduces an optional per-column format hook for cell display.
  - config/filters.js - registers Country and Primary hydrocarbon group filter dropdowns
  (Field status was already present).
  - ResourcesView.jsx - adds the new columns to
  the sort-status label map.

#### Country code → name display
  - constants/countries.js (new) - an ISO 3166-1 alpha-3 → alpha-2 code map plus getCountryName(), which resolves the conventional name at runtime via the platform Intl.DisplayNames API. No country-name strings are hard-coded and no new runtime dependency is added; unknown/blank values fall back to the raw value.
  - FilterBar.jsx / FilterDropdown.jsx - country filter options and active-filter chips show the conventional name, while the
  value sent to the API and used to clear chips remains the alpha-3 code (so the wire contract is unchanged).

#### Tests
  - constants/countries.test.js (new) - getCountryName mapping, unknown/empty fallbacks, case sensitivity, and ALPHA3_TO_ALPHA2 integrity (code shapes, seed codes present, no duplicate alpha-2).
  - ResourcesView.test.jsx - mock data/filter-options extended for the new fields; new tests for country rendering as a
  name (not code), country sorting, the new filter dropdowns, and that the country filter submits the code. 
  - Full frontend suite: 20 files / 178 tests  passing.

#### Notes / follow-ups (not in this PR)
  - The single-resource detail view still shows the raw country code - left for a follow-up since its value renderer is field-agnostic.
  - Country sorting orders by the stored alpha-3 code, not the display name (server-side sort + pagination). Sorting by
  display name would need a backend change or client-side sort.
  - package-lock.json was refreshed


jdhoffa (approved):
Awesome!! Thanks @jasonbrownrmi

PR #145, opened 2026-06-24
feat:remove query mixin

mbarlow12:
This PR is constrained only to "source" querying and lays the groundwork for reusable functions that can be applied to resource querying in a follow-up PR.

Removes the now-unnecessary `OGFieldQueryMixin` in favor of a dedicated query construction module  (`db/queries.py`).
Updates query workflow to be 2-phase: fetch ids, then use the id set to hydrate the response with a 2nd db call.
Adds minor optimizations to query construction
- only include colnames present in the query params
- drops json column handling  (owners & operators) since they can't participate in filtering or sorting


deployments/api/src/stitch/api/db/model/types.py
```diff
@@ -8,7 +8,11 @@
     .with_variant(sqlite.INTEGER(), "sqlite")
 )
 PORTABLE_JSON = JSON().with_variant(postgresql.JSONB(), "postgresql")
-PORTABLE_FLOAT = Float().with_variant(postgresql.DOUBLE_PRECISION(), "postgresql")
+PORTABLE_FLOAT = (
+    Float()
+    .with_variant(postgresql.DOUBLE_PRECISION(), "postgresql")
+    .with_variant(sqlite.REAL(), "sqlite")
```
AlexAxthelm, line 14: Thanks for catching the SQLite case

deployments/api/src/stitch/api/db/og_field_source_actions.py
```diff
@@ -159,10 +163,23 @@ async def query(
...
+    total = (
+        await session.scalar(select(func.count()).select_from(count_stmt.subquery()))
+        or 0
     )
-    return tuple(m.as_entity() for m in models), total
+
+    if not ids:
+        return (), total
+
+    headers = (
```
AlexAxthelm, line 179: I like this construction for clarity, and it's working great. Excited to see all the `awaits` in flight in Jaeger.

AlexAxthelm (approved):
Thanks for touching up my PR. Merging.

PR #146, opened 2026-06-24
refactor: consolidate query construction

mbarlow12:
## Summary

This branch reworks the OG-field query layer on top of the long-form (EAV) source values and tidies up how we pick the winning value for each field (priority, per-resource overrides, and licensing). Filtering/sorting/counting a list runs in SQL; the detail view and list hydration coalesce in Python. A parity test (`TestCoalescingEngineParity`) makes sure both paths pick the same winner.

## The list query (`db/queries.py`)

- `construct_base_query_statement` builds one shared CTE: active membership → resource → priority (+ any per-resource override) → source header → values, with licensing applied right there.
- `add_ranking` keeps the top-priority row per `(resource, colname)` with `ROW_NUMBER()` (works on SQLite, unlike `DISTINCT ON`), and `_add_pivot_columns` pivots only the fields a given query touches. Both list builders share this.
- The **source list** (`base_source_query`) selects just the header columns off the CTE, so the `GROUP BY` is valid on Postgres, and it filters by `source`.
- The **resource list** (`base_resource_query`) starts from the membership-derived set of resources and `LEFT JOIN`s the coalesced values onto it. A resource with no values you're licensed to see still shows up as a null shell on an unfiltered list, and drops out as soon as you filter on a field.

## Detail + hydration (`db/utils.py`, `db/model/resource.py`)

- `source_data_by_resource_id` is now a single query that mirrors the list CTE and returns `(source, effective priority)` rows; values selectin-load, and repointed resources are filtered out by `repointed_id IS NULL`.
- `coalesce_resources` returns one `OGFieldResource` per id — no more `repointed_ids` flag and no second pivot. It feeds both the detail view and list hydration (now inlined into `query`).
- Hydration takes a constant number of round-trips regardless of page size (no N+1).

## Public API

No change. `id` (filter) and `id`/`resource_id` (sort) map internally to `source_pk` on the source path and `resource_id` on the resource path; `SortableField` / `OGFieldFilterParams` are unchanged.

## Verification

- Full `stitch-api` suite passes (221); the parity test passes unchanged.
- Generated SQL checked on both SQLite and Postgres across id/filter/sort combinations.

## Heads-up for review

`main` solves this same area a different way (a single-pass `build_resource_list_cte`), so this branch will need to be reconciled with that before it merges.

## Jira

STIT-502


deployments/api/src/stitch/api/db/utils.py
```diff
@@ -45,36 +44,59 @@ async def resource_model_to_entity(
...
-    ]
-    rep_res: OGFieldResource | None = None
+    constituents = frozenset(
+        cm.id for cm in constituent_models if cm.id is not None and cm.id != model.id
+    )
+    repointed_to: int | None = None
     if model.repointed_id is not None:
         rep_model = await session.get(ResourceModel, model.repointed_id)
-        rep_res = rep_model.as_empty_entity() if rep_model else None
+        repointed_to = rep_model.id if rep_model else None
```
AlexAxthelm, line 58: NB: Is this supposed to handle the "return the repointed resource, rather than the literal requested one"?

deployments/api/tests/db/test_resource_actions.py
```diff
@@ -417,34 +428,6 @@ async def test_licensed_sources_none_returns_all(
         assert items[0].data.name == "RMI Name"
         assert items[0].provenance["name"] == "rmi"
 
-    @pytest.mark.anyio
-    async def test_licensed_sources_empty_returns_null_shells_for_all(
```
AlexAxthelm, line 421: NB: This test is okay to remove, because it's at the DB layer? there's still an equivilent one that the API returns nulls?

deployments/api/src/stitch/api/db/queries.py
```diff
@@ -84,62 +88,183 @@ def _participating_columns(params: OGFieldQueryParams) -> list[str]:
...
+        stmt = stmt.where(pivot_cte.c.source_pk == params.id)
+    return stmt.order_by(*_build_sort_clauses(pivot_cte, params, "source_pk"))
+
+
+def add_ranking(base_cte: CTE) -> Select[tuple[Any, ...]]:
+    cols = base_cte.c
+    ranked = (
+        select(base_cte)
+        .add_columns(
+            func.row_number()
```
AlexAxthelm, line 161: NB: I remember that on postgres (at least the older versions I used), there was some performance bump to using `DISTINCT ON` over `ROW_NUMBER` constructions. Can the ORM can handle routing to one structure in psql and then the other in sqlite? Not in anyway urgent, but something to think of in the future.

deployments/api/src/stitch/api/db/queries.py
```diff
@@ -84,62 +88,183 @@ def _participating_columns(params: OGFieldQueryParams) -> list[str]:
...
+        involved,
+        base_cte.c.colname,
+        lambda fn: getattr(base_cte.c, value_attr_for(fn)),
+    )
+    pivot_cte = pivot.cte("source_base")
+    stmt = select(pivot_cte.c.source_pk)
+    for cond in _build_field_conditions(pivot_cte, params):
+        stmt = stmt.where(cond)
+    # source is a source-path-only filter; the resource universe is source-ignoring.
+    stmt = stmt.where(pivot_cte.c.source.in_(list(dict.fromkeys(params.source))))
```
AlexAxthelm, line 150: I'm a little confused here, this looks like it's filtering on `params.source`, but elsewhere we're saying that we don't do that any more (which I think is the right behavior). is `params.source` a field we still need to keep around?

AlexAxthelm (approved):
Some questions for clarity, but nothing blocking

deployments/api/src/stitch/api/db/queries.py
mbarlow12: We'd aligned on not filtering by source for resources queries, but we handle the same params for querying sources. That's where it makes more sense to filer by source...we should probably get a UI built for that 😆

deployments/api/tests/db/test_resource_actions.py
mbarlow12: Opted to remove is since I think we discussed that a user with `()` permissions is a different error? That is, every user will have some kind of public permissions: `rmi`, `llm`, `gem`, I think?

deployments/api/src/stitch/api/db/queries.py
mbarlow12: I know both our implementations used the row number approach. There might've been a nested subquery or CTE that leverage it. Good candidate for a follow up tech debt/investigation issue.

I'm not sure if `DISTINCT ON` with `resource_id` & `colname` would beat the `where row_number = 1` overall. But definitely worth investigating.

deployments/api/src/stitch/api/db/queries.py
mbarlow12: Realized that we _do_ want to still allow filtering by `source` when querying the sources. Need to build a UI around that 😆

PR #152, opened 2026-06-26
Enable higher-resource containers

AlexAxthelm:
Some of our services require more memory than others, because for now they are loading a working dataset into memory, and then processing. This PR changes so that they are set as part of deploy, rather than requiring a manual tweak after the fact


jdhoffa (approved):
lgtm!

PR #154, opened 2026-06-30
Feat/added AGENTS.md

john:
## What changed

- Adds a project-scoped AGENTS.md to top level of repo.

## Verification

- Pull, ask the AI if this file is visible. Maybe ask it to do something crazy in direct defiance of what's in the file, see if it pushes back and references its instructions (do this in plan mode!)

## Human review

- This has worked well for me generally but could probably be specifically tuned for Stitch. My inclination is that once reviewed, if there are no issues, merge to main as is and iterate.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,4 +1,4 @@
-# Global Claude Code Instructions
+# Stitch AI Agent Instructions
 
 ## Operating Principles
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -98,7 +98,7 @@ Before adding or upgrading a dependency:
 * Optimize for first-time users.
 * When technical concepts must be exposed, explain them in user-focused language.
 * Prefer progressive disclosure: show only what most users need and reveal advanced details on demand.
-* Favor sensible defaults over requiring configuration.
+* Use sensible, discoverable defaults; reveal advanced configuration only when needed.
 
 ## Currentness
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,7 +44,7 @@ Before significant changes:
 * Use descriptive names.
 * Handle errors explicitly.
 * Do not silently ignore errors, exceptions, or failed promises.
-* Remove dead code when it is safe to do so.
+* Remove any dead code that your change orphaned. Do not delete pre-existing dead code, surface it in the summary instead.
 * Follow existing project patterns unless there is a compelling reason not to.
 * Optimize for future maintainers reading the code.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -78,7 +78,7 @@ Before adding or upgrading a dependency:
 * Do not commit unless explicitly asked.
 * Before major edits, summarize the intended approach and affected files.
 * After changes, provide a concise summary of what changed and why.
-* Never rewrite history, force-push, or delete branches unless explicitly instructed.
+* Never rewrite history, push, or delete branches unless explicitly instructed.
 
 ## Security
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,6 @@ Before adding or upgrading a dependency:
 ## User Experience
 
 * Prefer user-facing clarity and reduce cognitive load where possible.
-* Reduce cognitive load where possible.
 * Use plain language rather than technical jargon.
 * Favor interfaces that are easy for first-time users to understand.
 * Explain technical concepts in terms of user goals and outcomes.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,11 +21,12 @@
 Before significant changes:
 
 1. Understand the existing implementation.
-2. Explain the proposed approach.
-3. Identify meaningful tradeoffs.
-4. Ask clarifying questions when requirements are ambiguous.
-5. Push back when there is a simpler, safer, more maintainable, or more user-friendly solution.
-6. Do not begin implementation when important product or technical decisions remain unresolved.
+2. Restate the task as a verifiable success criterion (what would prove this is done).
+3. Explain the proposed approach.
+4. Identify meaningful tradeoffs.
+5. Ask clarifying questions when requirements are ambiguous.
+6. Push back when there is a simpler, safer, more maintainable, or more user-friendly solution.
+7. Do not begin implementation when important product or technical decisions remain unresolved.
 
 ## AI-Agent Guardrails
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,7 +33,7 @@ Before significant changes:
 * Do not make unrelated changes.
 * Do not perform broad refactors unless requested.
 * Do not replace working implementations with speculative rewrites.
-* Do not create new abstractions until there is a demonstrated need.
+* Do not create new abstractions until there explicit instruction.
 * Do not introduce new architecture unless explicitly justified.
 * Do not modify formatting in unrelated files.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -89,7 +89,7 @@ Before adding or upgrading a dependency:
 
 ## User Experience
 
-* Prefer clarity over flexibility.
+* Prefer user-facing clarity and reduce cognitive load where possible.
 * Reduce cognitive load where possible.
 * Use plain language rather than technical jargon.
 * Favor interfaces that are easy for first-time users to understand.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,108 @@
+# Global Claude Code Instructions
+
+## Operating Principles
+
+* Prefer clarity over flexibility.
+* Prefer simple interfaces over powerful interfaces.
+* Prefer root-cause fixes over symptom-level fixes.
+* Preserve existing behavior unless the task explicitly requires changing it.
+* Minimize exposure of implementation details.
+* Prefer small, safe, reversible changes.
+* Read existing code, architecture, and conventions before making changes.
+* Verify assumptions against the codebase rather than inferring them.
+* Understand the problem before proposing solutions.
+* When uncertain, state the uncertainty and choose the least risky path.
+* Favor maintainability and clarity over short-term cleverness.
+* Favor sensible defaults over requiring configuration.
+
+
+## Working Style
+
+Before significant changes:
+
+1. Understand the existing implementation.
+2. Explain the proposed approach.
+3. Identify meaningful tradeoffs.
+4. Ask clarifying questions when requirements are ambiguous.
+5. Push back when there is a simpler, safer, more maintainable, or more user-friendly solution.
+6. Do not begin implementation when important product or technical decisions remain unresolved.
+
+## AI-Agent Guardrails
+
+* Do not make unrelated changes.
+* Do not perform broad refactors unless requested.
+* Do not replace working implementations with speculative rewrites.
+* Do not create new abstractions until there is a demonstrated need.
+* Do not introduce new architecture unless explicitly justified.
+* Do not modify formatting in unrelated files.
+
+## Code Quality
+
+* Prefer boring, readable code over clever code.
+* Keep changes minimal and cohesive.
+* Use descriptive names.
+* Handle errors explicitly.
+* Do not silently ignore errors, exceptions, or failed promises.
+* Remove dead code when it is safe to do so.
+* Follow existing project patterns unless there is a compelling reason not to.
+* Optimize for future maintainers reading the code.
+
+For new files or files touched during implementation:
+
+* Run available linters and formatters before committing.
+* Resolve warnings that are directly related to the change.
+* Do not ignore failing checks without explanation.
+
+## Dependency Policy
+
+Before adding or upgrading a dependency:
+
+1. Check whether it is actively maintained.
+2. Prefer mature, widely adopted, well-documented libraries.
+3. Use the latest stable version unless compatibility requires otherwise.
+4. Avoid deprecated, abandoned, or niche packages.
+5. Check project runtime, framework, lockfile, and package manager compatibility.
+6. Explain dependency choices briefly.
+7. Do not introduce a dependency when the standard library or an existing dependency is sufficient.
+
+## Testing
+
+* Test changes proportionally to their risk.
+* Prefer existing test patterns and frameworks.
+* Add tests when introducing new behavior or fixing bugs where practical.
+* Do not claim something works without verification.
+
+## Git Workflow
+
+* Do not commit unless explicitly asked.
+* Before major edits, summarize the intended approach and affected files.
+* After changes, provide a concise summary of what changed and why.
+* Never rewrite history, force-push, or delete branches unless explicitly instructed.
+
+## Security
+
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,108 @@
+# Global Claude Code Instructions
```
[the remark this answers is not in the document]
jdhoffa: This is probably a good idea. Or just "Stitch Claude Code Instructions"

AGENTS.md
```diff
@@ -0,0 +1,108 @@
...
+5. Check project runtime, framework, lockfile, and package manager compatibility.
+6. Explain dependency choices briefly.
+7. Do not introduce a dependency when the standard library or an existing dependency is sufficient.
+
+## Testing
+
+* Test changes proportionally to their risk.
+* Prefer existing test patterns and frameworks.
+* Add tests when introducing new behavior or fixing bugs where practical.
+* Do not claim something works without verification.
```
jdhoffa, line 74: I have found success using this as a `CLAUDE.md` (with very few changes): https://github.com/multica-ai/andrej-karpathy-skills

The only substantive difference between the karpathy skills and the ones proposed in this PR is around "Goal-driven execution" which essentially asks the robot to conduct test driven development (https://github.com/multica-ai/andrej-karpathy-skills#4-goal-driven-execution)

@john thoughts about adopting that approach?

AGENTS.md
```diff
@@ -0,0 +1,108 @@
...
+## Working Style
+
+Before significant changes:
+
+1. Understand the existing implementation.
+2. Explain the proposed approach.
+3. Identify meaningful tradeoffs.
+4. Ask clarifying questions when requirements are ambiguous.
+5. Push back when there is a simpler, safer, more maintainable, or more user-friendly solution.
+6. Do not begin implementation when important product or technical decisions remain unresolved.
```
jdhoffa:
```suggestion
2. Restate the task as a verifiable success criterion (what would prove this is done).
3. Explain the proposed approach.
4. Identify meaningful tradeoffs.
5. Ask clarifying questions when requirements are ambiguous.
6. Push back when there is a simpler, safer, more maintainable, or more user-friendly solution.
7. Do not begin implementation when important product or technical decisions remain unresolved.
```

AGENTS.md
```diff
@@ -0,0 +1,108 @@
...
+* Do not modify formatting in unrelated files.
+
+## Code Quality
+
+* Prefer boring, readable code over clever code.
+* Keep changes minimal and cohesive.
+* Use descriptive names.
+* Handle errors explicitly.
+* Do not silently ignore errors, exceptions, or failed promises.
+* Remove dead code when it is safe to do so.
```
jdhoffa:
```suggestion
* Remove any dead code that your change orphaned. Do not delete pre-existing dead code, surface it in the summary instead.
```

AGENTS.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+## Working Style
+
+Before significant changes:
+
+1. Understand the existing implementation.
+2. Explain the proposed approach.
+3. Identify meaningful tradeoffs.
+4. Ask clarifying questions when requirements are ambiguous.
```
jdhoffa:
```suggestion
4. When multiple reasonable interpretations exist, present them and ask. Do not pick silently. When a requirement is unclear, name what is confusing before proposing a fix.
```

jdhoffa (commented):
Q: @john I suspect that you've used this `AGENTS.md` personally on other projects? How have you found that it handles small / trivial changes? Does it recognize that the change is tiny and skip to implementation, or does it arduously go through all instructions here? 

---

I've proposed a few changes that could better align this with https://github.com/multica-ai/andrej-karpathy-skills/blob/main/CLAUDE.md

AGENTS.md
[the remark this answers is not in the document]
john: I checked this and then forgot to edit the title, but turns out both Claude and Codex respect a general 'AGENTS.md', so I'll change to just 'Agent Guidance' or something, so it's tool-agnostic.

AGENTS.md
john: I think it depends how we ask it to do this and whether or not we want to put ourselves inside that loop. If it's ""Write tests for invalid inputs, then make them pass," and it does that itself, that's great (and what I've seen it do on its own). If it's "Agent writes tests for invalid inputs, human checks them out, then agent make them pass" I'm not keen on it, honestly I think we apply more scrutiny than we should. As Karpathy says, "These guidelines bias toward caution over speed," which if you're Andrej Karpathy and responsible for pre-training Claude, yes of course. Stakes are low for us and I'd rather prioritize speed.

But to that point if you'd prefer to add that and it gets this merged I'm totally fine with it, I defer to your experience. I presume it would merit its own section, perhaps above or below "Working Style"?

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+* Use descriptive names.
+* Handle errors explicitly.
+* Do not silently ignore errors, exceptions, or failed promises.
+* Remove any dead code that your change orphaned. Do not delete pre-existing dead code, surface it in the summary instead.
+* Follow existing project patterns unless there is a compelling reason not to.
+* Optimize for future maintainers reading the code.
+
+For new files or files touched during implementation:
+
+* Run available linters and formatters before committing.
```
AlexAxthelm, line 53: Tweak for stitch: explicit instructions to run the `make clean`, `make lint`, `make format` and `make check` targets, with reference to HACKING.md for tighter check loops (`make stitch-llm-test`),

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+5. Ask clarifying questions when requirements are ambiguous.
+6. Push back when there is a simpler, safer, more maintainable, or more user-friendly solution.
+7. Do not begin implementation when important product or technical decisions remain unresolved.
+
+## AI-Agent Guardrails
+
+* Do not make unrelated changes.
+* Do not perform broad refactors unless requested.
+* Do not replace working implementations with speculative rewrites.
+* Do not create new abstractions until there is a demonstrated need.
```
AlexAxthelm:
```suggestion
* Do not create new abstractions until there explicit instruction.
```

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+* Handle errors explicitly.
+* Do not silently ignore errors, exceptions, or failed promises.
+* Remove any dead code that your change orphaned. Do not delete pre-existing dead code, surface it in the summary instead.
+* Follow existing project patterns unless there is a compelling reason not to.
+* Optimize for future maintainers reading the code.
+
+For new files or files touched during implementation:
+
+* Run available linters and formatters before committing.
+* Resolve warnings that are directly related to the change.
```
AlexAxthelm, line 54: What about warnings that are not directly related, but arise as a result of refactor (noting that some of our CI fails on WARN)

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+
+Before adding or upgrading a dependency:
+
+1. Check whether it is actively maintained.
+2. Prefer mature, widely adopted, well-documented libraries.
+3. Use the latest stable version unless compatibility requires otherwise.
+4. Avoid deprecated, abandoned, or niche packages.
+5. Check project runtime, framework, lockfile, and package manager compatibility.
+6. Explain dependency choices briefly.
+7. Do not introduce a dependency when the standard library or an existing dependency is sufficient.
```
AlexAxthelm, line 67: Disagree here. There are times to bring in a dependency when stdlib can make a solution, often in times where managing the complexity of a home-built, or keeping it up to date is not worth it.

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Testing
+
+* Test changes proportionally to their risk.
```
AlexAxthelm, line 71: How do we identify risk here?

AGENTS.md
AlexAxthelm: A great deal of the work we've set up around non-local (shared PR previews and CI) dev work has an implict philosophy of "caution over speed". The robot is fast enough on its own. Keeping the option of tighter human review (the invalid test workflow you mentioned, or anything where progress is incremental over one-shot) should be considered valid, and perhaps encouraged.

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+* Prefer existing test patterns and frameworks.
+* Add tests when introducing new behavior or fixing bugs where practical.
+* Do not claim something works without verification.
+
+## Git Workflow
```
AlexAxthelm, line 76: Would it be useful to discuss attribution here? or is that all controlled through harnesses?

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+
+* Do not commit unless explicitly asked.
+* Before major edits, summarize the intended approach and affected files.
+* After changes, provide a concise summary of what changed and why.
+* Never rewrite history, force-push, or delete branches unless explicitly instructed.
```
AlexAxthelm:
```suggestion
* Never rewrite history, push, or delete branches unless explicitly instructed.
```

I have some concerns around including git operations in the discussion here, because I've gone to lengths locally to try to avoid letting the robots have any access to git, and I worry that having instructions like that might weaken or confuse.

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+* Never expose secrets or credentials.
+* Treat `.env*`, tokens, keys, and production configuration as sensitive.
+* Validate external inputs.
+* Prefer parameterized queries and framework-native escaping.
+* Call out security-sensitive changes explicitly.
+
+## User Experience
+
+* Prefer user-facing clarity and reduce cognitive load where possible.
+* Reduce cognitive load where possible.
```
AlexAxthelm: duplicate
```suggestion
```

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+* Use plain language rather than technical jargon.
+* Favor interfaces that are easy for first-time users to understand.
+* Explain technical concepts in terms of user goals and outcomes.
+* Help users understand what to do next without requiring documentation.
+* Optimize for first-time users.
```
AlexAxthelm, line 98: I don't think this the audience we want to optimize for with stitch?

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+* Never rewrite history, force-push, or delete branches unless explicitly instructed.
+
+## Security
+
+* Never expose secrets or credentials.
+* Treat `.env*`, tokens, keys, and production configuration as sensitive.
+* Validate external inputs.
+* Prefer parameterized queries and framework-native escaping.
+* Call out security-sensitive changes explicitly.
+
```
AlexAxthelm, line 90: We should have an accessibility section to encourage good practices there too

AlexAxthelm (commented):
I have some feedback on the particular document (below), but at the high level, I'm trying to understand the purpose of including this in Stitch?

This is fundamentally a document of opinions on how humans should include LLMs in the development process, by way of instructions to the robot. So these instructions are attempting to shape and homogenize the LLM process across the team.

I take issue with the necessity of that, since our team already uses the robots in different ways, both from each other, and depending on the work context. For example, my threads about front-end/JS implementations (which can be tested and checked locally) have a very different flow and cadence from when I'm working on CI plumbing (which can only be tested on machines to robot can't directly access)

That said, I think that having a file like this, but very tightly scoped to only stitch-specific things (robot, please look at HACKING and ARCHITECTURE) would be good to have

AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+## User Experience
```
mbarlow12, line 91: Is the "user" here the Claude Code/Codex user or the `stitch` end-user? If it's the latter, I think a tightly-scope `CLAUDE/AGENTS.md` located in `deployments/stitch-frontend` would get us more mileage when providing direction.

AGENTS.md
mbarlow12: What we lose out on if we take a no-git stance is that Claude Code is capable of creating and operating in worktrees and separate branches. We can also stop certain calls in the harness, e.g. adding the following to `settings.json` under the `PreToolUse` hooks prevents `git push...main/master`. Not bulletproof, but the agent would need to take a pretty adversarial approach to skirt it. 
```json
"type": "command",
"command": "CMD=$(jq -r '.tool_input.command'); if echo \"$CMD\" | grep -qE 'git[[:space:]]+push.*(main|master)'; then echo 'BLOCKED: Use feature branches, not direct push to main' >&2; exit 2; fi"
```
I haven't ever had issues with allowing a pretty large degree of freedom with `git`—I definitely have certain commands/patterns fully denied.

This does touch on what's better to have in context vs configured in the harness. The hook above provides that additional direction after preventing the call.

mbarlow12 (approved):
Generally, I think we can go ahead and merge then iterate on the details. We get the benefit of trying it out, checking agent results/behavior, and make targeted changes as we go.

**One minor fix (covered in #158):** Per [Anthropic docs](https://code.claude.com/docs/en/memory), Claude Code only reads `CLAUDE.md`, so a single `@AGENTS.md` line in `CLAUDE.md` gets all agents working harmoniously 👼 off the same context file.

Regarding the content itself, (NB) it's probably a bit too generic, so swapping in more `stitch`-specific guidance will be more useful in the long run.
- Name real commands: `make check` before pushing, tight loops like `make api-test`/`make stitch-llm-test`/`make frontend-test`/`make pkg-test-<name>`.
- Link `HACKING.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md` (via `@`) and remove overlapping/redundant content
- Trim the duplicated lines (the UX ones Copilot/Alex flagged).
- Later, consider per-area `CLAUDE.md` files (`deployments/api`, `deployments/stitch-frontend`, `packages/`) — Claude lazy-loads the nearest one, so a11y goes with the frontend, migrations with the api, etc.

As we refine, it's worth thinking about the aspects of our development practices that an agent **could not** learn from reading files in the repo.
- specific steps the agent should take before doing X. E.g.
	- "Run `make check` before committing changes."
	- "Search latest dependency documentation when responding in Plan mode."
- doc/sharepoint references or Jira guidance
	- "Relevant jira issues have been written to md files under `docs/jira/`"
	- "Use `acli jira ...` to surface relevant project task information"
- `tree` output or quick repo structure index (probably better suited for ARCHITECTURE.md)
- commands to check build CI/CD outputs/logs/status
- PR review steps/guidelines (though there are a lot of skills that handle this)
- proactive skill/tool use ("Use the context7 MCP to ground your responses in up-to-date documentation")

AGENTS.md
john: Don't get me wrong, I didn't mean to imply speed uber alles. More like, increasing relative to historical baseline, rather than reckless.

The PR previews and CI I'm super excited about, they feel like win-wins--good tooling and guardrails make it safer to go fast.

AGENTS.md
AlexAxthelm: Hmm. I think this is another one of those things where maybe we shouldn't standardize, and allow operators to find their own workflows.

I don't have any great philosophical objection to the robot having commit/push privileges, but I do object to it using my identity to do so.

PR #158, opened 2026-07-02
add CLAUDE.md that imports AGENTS.md

mbarlow12:
While there are claims that Claude Code can read `AGENTS.md`, official docs point to only `CLAUDE.md`, so this addition is to ensure that Claude Code will correctly import contents from `AGENTS.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md


AlexAxthelm (approved):
I think this makes some sense if we adopt #154, but noting that when I asked Claude, it said that it would fallback on AGENTS natively. 🤷

PR #159, opened 2026-07-03
Feat/see source values (STIT-493)

AlexAxthelm:
Add component to expand `FieldCard`s into a detailed view of the component sources in priority fallback order, with winner highlighted.

Adds endpoint to API to provide data for detailed view. Note that the current Resource Detail endpoint won't be sufficient for this after we introduce per-field priorities (upcoming PR).

Also note, the owner/operator fields aren't included in this at this time (more complex behavior)


deployments/stitch-frontend/src/hooks/useResources.js
```diff
@@ -343,6 +372,9 @@ export const useResourceDetail = USE_MOCK_DATA
 export const useSourceDetail = USE_MOCK_DATA
   ? useSourceDetailMock
   : useSourceDetailReal;
+export const useFieldSourceValues = USE_MOCK_DATA
+  ? useFieldSourceValuesMock
+  : useFieldSourceValuesReal;
 export const useMergeCandidates = USE_MOCK_DATA
   ? useMergeCandidatesMock
   : useMergeCandidatesReal;
```
mbarlow12, line 380: NB: I'm assuming that the "MOCK" mode is more for rapid FE development, is that right? This just smells a bit. We can mock dependencies in unit tests, but it's probably worth a backlog/tech debt ticket to consolidate so we're not declaring duplicated function variants.

deployments/api/src/stitch/api/db/og_field_resource_actions.py
```diff
@@ -127,6 +131,45 @@ async def get(
     )
 
 
+async def field_source_values(
```
mbarlow12, line 134: Do we need a separate endpoint? The `OGFieldDetailView` already has a full list of sources, sorted in priority order. Could we drop the extra API call & endpoint if we use that data?

deployments/stitch-frontend/src/pages/ResourceDetailPage.jsx
mbarlow12, line 539: We're waiting here to load the full resource view as it is. Why not fetch data per attribute in pure JS?

mbarlow12 (commented):
The main thing  I noticed is that we're already populating a full list of source data when a user hits the `{id}/detail` endpoint, responding with `OGFieldDetailView`. Its `source_data` attribute has all the data we need to display the list of values for a given attribute with the exception of the actual `priority` number, but the sources are already sorted in priority order. Unless we need to display "1/2/3/4/...", I think the dropdown could be displayed in pure JS without a second API call.

deployments/stitch-frontend/src/hooks/useResources.js
AlexAxthelm: @jasonbrownrmi introduced the MOCK paradigm, I believe to decouple frontend development from the need to have an API running at all times

deployments/api/src/stitch/api/db/og_field_resource_actions.py
AlexAxthelm: We need a finer grained endpoint, since we'll need to move away from the idea that a resource only has a single set of priorities, and to the idea that each field has its own set of priorities (so the priorities for "Basin" and "Local name" may be differently ordered).

To that end, I think dropping the full list of sources from the OGFieldDetailView is probably a good call moving forward (or at least don't imply that it's an ordered set), and if we need source details, we can fetch those as needed.

deployments/stitch-frontend/src/pages/ResourceDetailPage.jsx
AlexAxthelm: I think this is related to the other comment about needing a finer grain on the priority details, and possibly dropping some detail from the details endpoint

deployments/api/src/stitch/api/db/og_field_resource_actions.py
mbarlow12: If we drop the full list of sources, then the Detail view becomes just the ListItemView and forces us to make calls for every field attribute the user wants to view. Any other service will need to do the same to get the full, equivalent data for a Field.

mbarlow12 (approved):

PR #166, opened 2026-07-09
Curate AGENTS.md

mbarlow12:
Replace the generic AGENTS.md with curated, repo-specific guidance. The CLAUDE.md → @AGENTS.md bridge is already on this branch, so the diff is AGENTS.md only.

This was almost entirely generated by Claude with some minor edits after the fact.

I first ran the `/init` slash command to have CC review the repo in detail and create a CLAUDE.md. Then, I went through an interactive merge process/conversation to combine the contents of the 2 files while still keeping under the 200 line recommended limit. I use the official [claude-md-management plugin](https://github.com/anthropics/claude-plugins-official/blob/main/plugins/claude-md-management/README.md) that has a `claude-md-improvement` skill which has been useful for me across a number of project.

There's a shift toward orienting agents to the repo with common commands, structure/layout, notable gotchas, etc...
> [!IMPORTANT]
> With the added specifics, it'll be more important to keep AGENTS.md & supporting docs updated with relevant content (e.g. the source key list).

The hope is that this should offer decent guidance/guardrails on working in `stitch` without overly dictating specific workflows or patterns. And it should pair well with user-level `~/.claude/CLAUDE.md` or local overrides (`CLAUDE.local.md`).


AGENTS.md
```diff
@@ -1,108 +1,127 @@
...
+
+Pointers: [`HACKING.md`](./HACKING.md) (setup & day-to-day workflow) ·
+[`ARCHITECTURE.md`](./ARCHITECTURE.md) · [`CONTRIBUTING.md`](./CONTRIBUTING.md) ·
+[`deployments/CI_DEPLOYMENTS.md`](./deployments/CI_DEPLOYMENTS.md) (deploy pipeline) ·
+[`deployments/PERFORMANCE.md`](./deployments/PERFORMANCE.md) (query-layer performance).
+
+## How to work here
+
+- Read the surrounding code and conventions first; verify against the codebase rather than
+  inferring.
```
AlexAxthelm:
```suggestion
- Read the surrounding code and conventions first; verify against the codebase rather than inferring.
```

@mbarlow12 I wasn't aware of a 200 line guidance on files like this (probably because I'm not currently using them). Do they also have guidances on length? Or are these breaks coming from our formatters?

AGENTS.md
```diff
@@ -1,108 +1,127 @@
...
+- `make api-dev` / `make frontend-dev` / `make dev-docker` — the three dev entrypoints (see `HACKING.md` for which to pick).
+- `make uv-sync-dev` — sync the whole workspace (`uv sync --group dev --all-packages`).
+- `make clean` — remove build artifacts, caches, and dev docker volumes.
+
+Single-package / single-test loops (the aggregate `make test` runs everything and is slow — see
+`HACKING.md` for tight loops):
+
+- `make api-test`, `make entity-linkage-test`, `make stitch-llm-test`, `make seed-test`, `make frontend-test`, `make pkg-test-<name>` — per-target suites.
+- One file or test: `uv run --package <pkg> pytest <path>[::test_name] -x`, e.g.
+  `uv run --package stitch-api pytest deployments/api/tests/db/test_query.py -x`.
```
AlexAxthelm, line 56:
```suggestion
  `uv run --package stitch-api pytest deployments/api/tests/db/test_query.py -x`.
  
While the per-target suites and `uv` invoked tests are good for iterating quickly, note that `make check` and the higher-level targets should pass cleanly before work is considered complete.
```

AGENTS.md
```diff
@@ -1,108 +1,127 @@
...
+- The persistence model is EAV-style (see `deployments/api/src/stitch/api/db/model/`:
+  `resource.py`, `membership.py`, `oil_gas_field_source*.py`, the `*_priority.py` tables) with
+  a human-review flow via `merge_candidate.py`.
+
+## Supporting services & deployment direction
+
+`stitch-api` is the intended central broker / control plane — it owns auth, the domain model,
+and the canonical data. `entity-linkage`, `stitch-llm`, and the ETL app are supporting
+capabilities.
+
```
AlexAxthelm, line 93:
```suggestion
Note that the ETL processes are defined in a separate repo, but inherit interfaces that are defined here. the `seed` deployment provides a public approximation.

```

AGENTS.md
```diff
@@ -1,108 +1,127 @@
...
+  (`ROW_NUMBER()`). The API `conftest.py` skips creating tables marked
+  `info={"is_view": True}` — anything that must exist in the SQLite test DB must not carry
+  that marker.
+- **The frontend is configured at runtime**, not at build time, via
+  `deployments/stitch-frontend/public/config.json` (API/entity-linkage/LLM/ETL URLs, Auth0).
+  CI injects these per deploy lane.
+
+## CI gates that will fail a PR
+
+- A **forbidden-patterns** scan (`.github/workflows/admin-forbidden-patterns.yml`) blocks
```
AlexAxthelm, line 108: Open question for the group, do we want to add `forbidden-patterns` to `make check` or something similar?

AlexAxthelm (commented):
Big fan of this! Could not have written better myself. Some small notes from my experiences with the robots. Thanks.

AGENTS.md
```diff
@@ -1,108 +1,127 @@
...
-## Currentness
-
-For packages, APIs, frameworks, cloud services, security topics, and operational procedures:
-
-* Verify current documentation rather than relying solely on memory.
-* Prefer official documentation, release notes, maintainers, and primary sources.
+# Stitch — Agent Guidance
+
+Stitch integrates diverse oil & gas asset datasets, applies AI-driven enrichment with human
+review, and delivers curated, trustworthy data. This file captures the things that aren't
```
john, line 4: This is taken straight from the README, so it will be in context already. Should we reserve AGENTS.md for instructions and information it won't otherwise get?

AGENTS.md
```diff
@@ -1,108 +1,127 @@
...
+obvious from reading one file.
+
+Pointers: [`HACKING.md`](./HACKING.md) (setup & day-to-day workflow) ·
+[`ARCHITECTURE.md`](./ARCHITECTURE.md) · [`CONTRIBUTING.md`](./CONTRIBUTING.md) ·
+[`deployments/CI_DEPLOYMENTS.md`](./deployments/CI_DEPLOYMENTS.md) (deploy pipeline) ·
+[`deployments/PERFORMANCE.md`](./deployments/PERFORMANCE.md) (query-layer performance).
```
john, line 10: Same comment as above--these are linked from the README, so why are they necessary here? If we're concerned README material will fall out of context, we might be able to be more economical in AGENTS if there's just a line saying "Read the README"

Skimmed the rest and that will be my general comment--most of what's been added already exists in top-level *.md files. I think we should do some research about how agents treat existing docs vs AGENTS, so we can take advantage of this new file while reducing duplication as much as possible. It's not just that adding that material bumps other material, given suggested length limits for AGENTS, it also means there's no longer one source of truth, we'll have to keep two files updated if anything changes.

jdhoffa: Addressed in https://github.com/RMI/stitch/pull/195

PR #170, opened 2026-07-17
Refactor/coalesce

AlexAxthelm:
Addresses technical debt: coalescing logic lives in the DB (SQL is the single coalescer), with no source-coalescing logic left in Python. Initial performance testing (PR #171) indicates comparable performance to baseline.

**Empty strings (behavior change, now a DB invariant).** Previously the Python and SQL coalescers diverged on empty strings — one treated `""` as a winnable value, the other filtered it. Rather than reconcile two behaviors, empty text is no longer persistable: the write path drops empty-string attributes, and a DB `CHECK (value_text <> '')` (`ck_source_value_text_nonempty`, migration `790b2078d706`) enforces it. NULL/absent is now the single "unset" sentinel, so the runtime `!= ""` filters are gone.

**Detail hydration is a single query.** `resource_model_to_entity` builds the coalesced view + provenance (winner rows) and the raw `source_data` (grouped by source) from one ranked query, instead of a separate winner query + source-listing query.

**Merge preview dropped (breaking).** Removes `GET .../merge-candidates/{id}/preview` and `OGFieldMergePreviewView`; this will be unused in upcoming merge-review changes. The review page shows a "Merge preview is unavailable." message in the meantime.

Migrations are forward-only (irreversible downgrade), per the repo convention.


deployments/api/src/stitch/api/db/queries.py
```diff
@@ -174,6 +181,31 @@ def add_ranking(base_cte: CTE) -> Select[tuple[Any, ...]]:
...
+    winners = add_ranking(base).cte("coalesced_winners")
+    return select(
+        winners.c.resource_id,
+        winners.c.colname,
+        winners.c.value_text,
+        winners.c.value_num,
+        winners.c.value_json,
+        winners.c.source,
+        winners.c.source_pk,
+    ).where(winners.c.resource_id.in_(list(dict.fromkeys(resource_ids))))
```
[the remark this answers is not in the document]
AlexAxthelm: https://github.com/RMI/stitch/pull/170/commits/6af7d6508596dee7e68f3ab1b7872bba659f62c5

deployments/api/tests/db/test_resource_actions.py
```diff
@@ -287,6 +287,41 @@ async def test_no_redaction_uses_priority_coalesced_scalar_fields(
...
 
+    @pytest.mark.anyio
+    async def test_empty_string_value_loses_to_lower_priority_nonempty(
+        self,
+        seeded_integration_session: AsyncSession,
+        test_user: User,
+    ):
+        # Highest-priority source (rmi) carries an empty-string basin; coalescing
+        # must fall through to the lower-priority non-empty value instead of
+        # letting the empty string win.
```
mbarlow12: I think this should be a constraint at the DB layer instead of handled in queries/coalescing. Basically, we shouldn't allow empty strings when persisting to the DB & if one is found that's an error. That way NULL/None **always** means its unset and we don't need subsequent checks for empty strings.

deployments/api/src/stitch/api/db/utils.py
```diff
@@ -46,6 +52,13 @@ async def resource_model_to_entity(
 ) -> OGFieldResource:
     base = (await coalesce_resources(session, [model.id], licensed_sources))[model.id]
 
+    # coalesce_resources builds only the coalesced view + provenance (SQL-side).
+    # Detail views also expose the raw per-source rows, so fetch them here.
+    rows_by_id = await ResourceModel.source_data_by_resource_id(
+        session, [model.id], licensed_sources
+    )
+    sources = [src for src, _ in rows_by_id.get(model.id, [])]
+
```
mbarlow12: `resource_model_to_entity` now calls `source_data_by_resource_id` here and `coalesced_winner_rows`  (via `coalesce_resources`). 2 queries to render the entity where there was originally one. The windowed query is redundant work, and doesn't address the underlying data issue that `NULL != ""`.

deployments/api/src/stitch/api/db/utils.py
```diff
@@ -67,82 +84,83 @@ async def coalesce_resources(
...
+    if ids:
+        rows = await session.execute(coalesced_winner_rows(ids, licensed_sources))
+        for rid, colname, value_text, value_num, value_json, source, source_pk in rows:
+            value = materialize_value(
+                colname,
+                value_text=value_text,
+                value_num=value_num,
+                value_json=value_json,
+            )
+            provenance_by_id[rid][colname] = (value, source, source_pk)
```
mbarlow12: My concern here is less about actual performance and more about executing redundant work. The windowed query was built to perform user requested searches over the attributes, so once there's a set of `resource_ids`, that SQL doesn't serve the use case well.

mbarlow12 (changes requested):
I'm a little confused on what the observed divergence was. But, this is subtly introducing new behavior. Originally, we were treating `""` as valid input and it would win in coalescing, implying that `NULL != ""`. This PR essentially flips that, and we treat empty strings as `NULL`.

The trouble is that it adds redundant runtime work and a lot of defensive checks when we could maintain an invariant in the DB: `CHECK(value_text <> '')` and optionally coerce `""` into `NULL` on writes. We can add non-empty constraints on the pydantic fields, too, if we want. This prevents empty strings from being persisted, and allows us to keep the original query and coalescing logic.

deployments/api/src/stitch/api/db/utils.py
AlexAxthelm: Fixed in `7280e41`. `resource_model_to_entity` now runs a single query (`coalesced_candidate_rows`) and derives both projections from it. I ended up pulling some logic that I had in #174 into here instead, but you were right, this is a better way to handle it. Basically, we're now grabbing the rankings along with the source details from the DB, and when constructing the winners, we just filter to the best ranked, when constructing the sources, we drop rankings, and group by source. same (single) query, just different presentations.

Note that I'm keeping the lazy-load field-sources endpoint for now, to keep this PR somewhat constrained, but flagged https://rmi1.atlassian.net/browse/STIT-594 to bring the full field-source objects into the detail view

deployments/api/src/stitch/api/db/utils.py
AlexAxthelm: I think the tradeoff of a maintaining a single coalescer is worth allowing some redundant work. 

That said, there is some partial work towards this comment in the PR: the query is narrowed when it can be, so it partitions a single resources rows, rather than the whole table, and as mentioned above, we're now single-tripping to the DB, and reading the rankings from there, rather than making two hops.

Your deeper point — that for a known id-set a window function is heavier than strictly necessary is spot on though. The fuller fix (fold the per-field rankings into the detail payload, derive by-source client-side, retire the lazy field-source read endpoint) is tracked as https://rmi1.atlassian.net/browse/STIT-594

deployments/api/tests/db/test_resource_actions.py
AlexAxthelm: Spot on. thanks. Feels obvious in hindsight.

https://github.com/RMI/stitch/pull/170/commits/ffe27a9f8cf01f7821de9bbab8e2a79c07545e43

deployments/api/src/stitch/api/db/queries.py
```diff
@@ -153,9 +162,24 @@ def base_source_query(
...
+    (``coalesced_candidate_rows``) so the two can't drift. ``rn == 1`` is the
+    winner within each ``(resource_id, colname)`` partition.
+
+    MERGE(174): 174 introduces this same ``_ranked`` split, but with tiered
+    per-field override ranking (``override_priority`` NULLS LAST, then
+    ``default_priority``, then source/source_pk). On merge keep 174's richer
+    ``order_by`` -- this branch differs there only because per-field overrides
+    don't exist here yet. BUT drop 174's ``.where(value_text != "")`` empty-text
+    filter: this branch makes empty strings impossible at write time + a DB
+    CHECK (see ``model.oil_gas_field_source_value``), so that filter is now dead.
```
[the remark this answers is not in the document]
AlexAxthelm: would rather keep this for now, 174 coming soon.

deployments/api/src/stitch/api/db/model/oil_gas_field_source_value.py
```diff
@@ -139,6 +139,13 @@ class OilGasFieldSourceValueModel(Base):
             "colname IN (" + ", ".join(f"'{n}'" for n in ATTRIBUTE_NAMES) + ")",
             name="ck_source_value_colname",
         ),
+        # Empty text is never persisted: NULL/absent is the single "unset"
+        # sentinel, so coalescing never has to treat "" as a candidate value.
+        # (NULL passes -- the check only rejects the empty string.)
+        CheckConstraint(
+            "value_text <> ''",
+            name="ck_source_value_text_nonempty",
+        ),
```
mbarlow12, line 148: Worth adding `json`constraints as well?
```python
CheckConstraint(
            """
            data IS NULL
            OR (
                jsonb_typeof(data) IN ('object', 'array')
                AND data NOT IN ('{}'::jsonb, '[]'::jsonb)
            )
            """,
            name="ck_source_value_json_nonempty",
        )
```
I realize this isn't supported by `sqlite`, but it seems like we've moved to the point where it's not carrying its weight. Hence [STIT-603](https://rmi1.atlassian.net/browse/STIT-603)

mbarlow12 (approved):
One NB/optional addition. LGTM.

deployments/api/src/stitch/api/db/model/oil_gas_field_source_value.py
mbarlow12: Created [STIT-613](https://rmi1.atlassian.net/browse/STIT-613) instead.

PR #174, opened 2026-07-20
Feat/re prioritize field (STIT-494)

AlexAxthelm:
supersedes #164 

Lets a `resource:write` curator reorder sources for each field in a resource, and persist in database (changing ordering for all users).

Builds on sql-only coalescing logic (#170)

* Updates `og_field_resource_source_priority` table in DB to include field name (per field reordering rather than overall)
* introduces a tiered ranking where explicitly ordered sources (have override) rank above sources with no override
* Adds appropriate endpoints: `PUT /oil-gas-fields/{id}/fields/{field}/sources/priority`
* includes draft frontend UI elements


deployments/api/src/stitch/api/routers/oil_gas_fields.py
```diff
@@ -277,6 +278,48 @@ async def get_field_source_values(
...
+    response_model=list[OGFieldSourceValueView],
+    dependencies=[Depends(require_permissions(RESOURCE_WRITE))],
+)
+async def set_field_source_priority(
+    *,
+    uow: UnitOfWorkDep,
+    user: CurrentUser,
+    claims: Claims,
+    id: int,
+    field: str,
```
mbarlow12: NB: we could further constrain this to a `Literal[...]` that's maintained in `stitch-ogsi`.

deployments/api/src/stitch/api/db/og_field_resource_actions.py
```diff
@@ -151,23 +163,108 @@ async def field_source_values(
...
+        )
+    if len(set(ordered_source_pks)) != len(ordered_source_pks):
+        raise InvalidActionError("ordered_source_pks contains duplicate ids.")
+
+    # Eligibility + current effective order come from the same ranked read the GET
+    # endpoint uses: licensed sources with a non-empty value for the field,
+    # winner-first.
+    rows = (
+        await session.execute(field_source_candidates(id, field, licensed_sources))
+    ).all()
```
mbarlow12, line 241: I think we need to pass all possible `licensed_sources` to this query. As it stands, if 2 users with `resource:write` permissions have different `source:read:*` permissions, they'll silently override the settings.

E.g.
- Curator A (all licenses) curates name as [ccr, rmi, wm, gem] → resource name is "CCR Name".
- Curator B, licensed for gem, rmi, & wm only, sees a three-row panel and sets priorities to `wm`, `rmi`, `gem` (the result from `field_source_candidates` will match the requested set)
- Result: we call `delete` on resource_id + colname, so `ccr` gets silently deleted.

We should either:
- loudly raise if a user tries to adjust source priorities where there are existing overrides for sources they can't read
- only change the priorities for the user's licensed sources, so in the above example, Curator B's request would only rearrange rmi, wm, and gem, resulting in the db storing [ccr, wm, rmi, gem]. Said another way, user's cannot change source priorities for sources they can't read.

deployments/stitch-frontend/src/components/ResourceFieldCard.jsx
```diff
@@ -20,24 +41,175 @@ function SourceValueRow({ source, value, sourceId, isWinner }) {
...
+    try {
+      const fetcher = createAuthenticatedFetcher(
+        config,
+        getAccessTokenSilently,
+      );
+      await updateFieldSourcePriority(
+        config,
+        resourceId,
+        fieldKey,
+        workingOrder,
```
mbarlow12, line 297: Not big, but it's _technically_ possible for `workingOrder` to become stale if `sources` somehow fetches & updates while the panel is open.

mbarlow12 (changes requested):
Big one is a sneaky potential for users with differing read permissions to overwrite priorities to all sources on a resource + attribute. See the second comment.

Other notes:
- do we need/want some kind of "reset to default" that deletes the rows from the overrides table?
- more integration tests for the endpoint? might be able to test the permissions bug explicitly

deployments/api/src/stitch/api/db/og_field_resource_actions.py
AlexAxthelm: [f7786fd](https://github.com/RMI/stitch/pull/174/commits/f7786fd7b3b73984cfaa65347cd1788f099d6f01)

check for all read permissions, to avoid complicated debugging behavior

deployments/stitch-frontend/src/components/ResourceFieldCard.jsx
AlexAxthelm: [7fc673f](https://github.com/RMI/stitch/pull/174/commits/7fc673f7c5add01f7a5331450174a7c68209eedd)

deployments/api/src/stitch/api/routers/oil_gas_fields.py
AlexAxthelm: [109be9c](https://github.com/RMI/stitch/pull/174/commits/109be9cf4a1a3795a6290983e976ba539a728068)

deployments/stitch-frontend/src/components/ResourceFieldCard.jsx
```diff
@@ -154,33 +215,152 @@ function FieldSourcesPanel({
...
+    !arraysEqual(
+      [...workingOrder].sort((a, b) => a - b),
+      [...originalOrder].sort((a, b) => a - b),
+    );
+
+  function beginEdit() {
+    setWorkingOrder(originalOrder);
+    setSaveError("");
+    setIsEditing(true);
+  }
```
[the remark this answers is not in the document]
AlexAxthelm: [df652db](https://github.com/RMI/stitch/pull/174/commits/df652db8e65ff4bc33cc182a34e1ef55b5d534fd)

mbarlow12 (approved):
Looks good 👍🏼

PR #183, opened 2026-07-21
STIT-574 side by side comparison

jasonbrownrmi:
## Summary

Replaces the merge candidate review page's "merged preview" panel with a side-by-side comparison table of the source resources, so reviewers can see exactly where sources agree and disagree before deciding. Once a candidate has actually been merged, the comparison is replaced with a view of the merged resource itself, and source resource ids are now clickable links to their detail pages.

- Adds `MergeSourceComparison`, which fetches each source resource and renders a per-field comparison table, flagging each row as `match`, `differs`, or empty (`deployments/stitch-frontend/src/components/MergeSourceComparison.jsx`).
- Adds shared `FIELD_META` constants describing which fields to compare, their labels, and section grouping (`deployments/stitch-frontend/src/constants/fieldMeta.js`).
- Adds `getRowStatus` comparison logic (`deployments/stitch-frontend/src/utils/mergeComparison.js`); string values are compared case-insensitively so differences in casing alone don't get flagged as a mismatch.
- Adds `MergedResourceView`, a labeled single-column view (reusing the same `FIELD_META` fields as the comparison table) that replaces the comparison once a candidate's `merged_resource_id` is set — i.e. once the merge has actually run.
- Links each source resource id in the candidate summary to its detail page (`/oil-gas-fields/:id`) instead of showing plain, unlinked ids.
- Removes the old `PreviewPanel` (merged-preview) and collapsible `SourceResources` section from `MergeCandidateReviewPage.jsx` in favor of the components above.
- Keeps the decision panel mounted and interactive while candidate detail is loading/refreshing, falling back to the already-loaded queue list item so review actions don't dead-click; shows a "these details may be out of date" banner if the detail refresh fails instead of hiding the panel.
- Renders loading state as placeholder/skeleton cells (with a spinner) instead of blocking the whole panel, so headers and layout stay stable while sources load.

## Test plan

- [x] `npx vitest run` — all 220 tests passing (`mergeComparison.test.js`, `MergeSourceComparison.test.jsx`, `MergedResourceView.test.jsx`, `MergeCandidateReviewPage.test.jsx`, and the rest of the suite)
- [x] `eslint` and `prettier --check` clean on all touched files
- [x] Manually review a pending merge candidate in the app and confirm the comparison table renders correctly for matching, differing, and empty fields
- [x] Manually review an approved (merged) candidate and confirm the merged resource view renders in place of the comparison
- [ ] Click a source resource id link and confirm it navigates to the correct detail page

## Notes for reviewers
- Not related to this PR, but surfaced during review: `useMergeCandidatePreview` (and its mock/real variants in `hooks/useResources.js`), plus `resourceQueries.mergeCandidatePreview`/`resourceKeys.preview` in `queries/resources.js` and `getMergeCandidatePreview` in `queries/api.js`, are now unused — they supported the old merged-preview panel this PR removes. Left in place pending a decision on whether a "preview before merging" feature is still wanted.


.gitignore
```diff
@@ -220,3 +220,6 @@ __marimo__/
 
 # macOS
 **/.DS_Store
+
+# Local design notes and plans
+/docs/
```
AlexAxthelm, line 225: Do we actually want this to be ignored? cc @jdhoffa @mbarlow12

deployments/stitch-frontend/src/utils/mergeComparison.js
AlexAxthelm, line 1: @jasonbrownrmi Is this going to be removed if we add the compare object to the merge candidate (PR #173)?

AlexAxthelm (approved):
Overall looks good

.gitignore
jdhoffa: Fine to ignore this. If we eventually do want to use this to document anything we can always unignore it.

deployments/stitch-frontend/src/utils/mergeComparison.js
jdhoffa: Let's follow up on this in https://rmi1.atlassian.net/browse/STIT-596

jdhoffa (changes requested):
@jasonbrownrmi confirm that clicking on the relevant links to the merge candidates link to the appropriate `ResourceDetail` page, however there is no option to click on the newly merged Resource number. 

Any chance we could fix that? 

Image

jdhoffa (approved):
Alex has addressed my comment. LGTM!

PR #184, opened 2026-07-21
feat: STIT-575 show resolved candidate names instead of raw ids in merge review

jasonbrownrmi:
## Summary

The merge candidate queue and detail panel previously identified every candidate only by its own database id (`Candidate #11`) and listed its source/merged resource ids as plain text. This PR resolves a real display name for each candidate from its source resources, and tightens up how status is communicated.

- Adds `pickCandidateName`, a pure function that picks a candidate's display name from its source resources' `name` field using the same source-priority rules as backend coalescing (`rmi` > `gem` > `wm` > `llm`; ties broken by resource order) — see `deployments/stitch-frontend/src/utils/mergeCandidateName.js`.
- Adds `SOURCE_PRIORITY` to `constants/sourceMeta.js`, mirroring the backend's `SRC_PRIORITY` (`deployments/api/src/stitch/api/coalesce.py`).
- Adds `useMergeSourceDetails`, extracted from `MergeSourceComparison`'s inline query so the fetch of a candidate's source resources is shared (same cache entry) between the comparison table and the new name lookup — no duplicate network requests for the same candidate.
- Adds `useMergeCandidateName`, combining the two above into a single hook.
- The queue list and detail panel heading now show the resolved name, falling back to `Candidate #{id}` if no source has a name.
- Source resource ids and the merged resource id are no longer shown as plain list text in the queue; they're still visible in the detail panel's existing facts, and available via a hover tooltip on each queue row.
- The status badge now reads **CANDIDATE** instead of **PENDING** (same styling/color), since "pending" reads oddly once the item has a real name — APPROVED/DENIED are unchanged. This replaces an earlier iteration that used two separate badges.
- Fixed a layout bug where the name + badge row would wrap, leaving the badge stranded at the start of the next line; the row no longer wraps as a whole — the name wraps within its own space while the badge stays pinned to the right. Also slightly reduced the badge's padding.

## Test plan

- [x] `npx vitest run` — all 230 tests passing, including new coverage for `pickCandidateName` (priority selection, tie-breaking, missing/empty names, unrecognized sources) and the queue/detail name + badge behavior
- [x] `eslint` and `prettier --check` clean on all touched files
- [ ] Manually review the merge queue and confirm names resolve correctly across a few real candidates, and that long names wrap without displacing the badge
- [ ] Confirm the hover tooltip on a queue row shows the expected source/merged resource ids

## Notes for reviewers

- This branch is stacked on `STIT-574-side-by-side-comparison` (not yet merged); the diff above is scoped to just this branch's own commits.


deployments/stitch-frontend/src/utils/mergeCandidateName.js
AlexAxthelm, line 1: similar to #183, I think that if we adopt #173, we can let the api deliver its "best guess" name (`candidate.compare.name.values[0]`), which would let us avoid having to have coalescing logic in JS.

jasonbrownrmi: I agree, that makes sense to me.

jdhoffa (approved):

PR #185, opened 2026-07-21
feat: integrate C&C reservoirs

mbarlow12:
Integrates the C&C Reservoirs (CCR) dataset end-to-end: adds the `ccr` source key through OGSI, coalesce priority + the source-priority migration, the `source:read:ccr` permission, and a CCR trigger panel on the ETL page.

Also adds  single constant for python source priorities. May be overridden soon, but it cleans up a bit of redundant code.

Companion ETL change (turns on CCR posting) is `RMI/stitch-etl-poc` [#16](https://github.com/RMI/stitch-etl-poc/pull/16).

Running live on staging via the demo PR #192 if you want to poke at it.

STIT-540


deployments/api/alembic/versions/e5dfcbfc32e3_add_ccr_source_priority.py
```diff
@@ -0,0 +1,38 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "e5dfcbfc32e3"
+down_revision = "f3fb36006ce6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Renumber to rmi=1, gem=2, wm=3, ccr=4, llm=5. Move llm to the (free)
```
AlexAxthelm, line 21: Shouldn't WM come before GEM?

deployments/api/src/stitch/api/db/model/og_field_source_priority.py
```diff
@@ -18,7 +18,8 @@ class OGFieldSourcePriority(Base):
     {"source": "rmi", "priority": 1},
     {"source": "gem", "priority": 2},
     {"source": "wm", "priority": 3},
```
AlexAxthelm: Non-blocking for this PR, but similar to above, shouldn't the GEM and WM be switched here?

Also is there a good way to check/test that the DB table is in alignment with this object? It seems like there's a possibility of drift between them.

deployments/api/tests/test_coalesce.py
```diff
@@ -0,0 +1,47 @@
+"""Coalescing priority tests for the CCR source.
+
+CCR must rank between ``wm`` and ``llm``:
+    rmi (1) > gem (2) > wm (3) > ccr (4) > llm (5)
+so a higher-priority source's value wins a contested field.
```
AlexAxthelm: NB: Is this test still relevant if we adopt #170?

packages/stitch-models/src/stitch/models/types.py
```diff
@@ -8,7 +8,7 @@
     None | bool | int | float | str | list["JsonValue"] | dict[str, "JsonValue"]
 )
 
-Year = Annotated[int, Field(ge=1800, le=2100)]
+Year = Annotated[int, Field(ge=1700, le=2100)]
```
AlexAxthelm, line 11: Possible Blocker: Why are we changing this limit? If this is changing, shouldn't we be updating some tests somewhere?

AlexAxthelm (approved):
tentatively approved, pending discussion of the `Year` comment (1700 vs 1800)

deployments/api/alembic/versions/e5dfcbfc32e3_add_ccr_source_priority.py
mbarlow12: Yup...I'm making this a single constant somewhere.

packages/stitch-models/src/stitch/models/types.py
mbarlow12: I could either drop the handful of values this hits or expand here. It's only 2 fields, but apparently it's legit (read through AI surfaced references)...
- [oil sands magazine](https://www.oilsandsmagazine.com/technical/history-of-alberta-oil-sands-development)
- [wikipedia](https://en.wikipedia.org/wiki/Athabasca_oil_sands)
- [alberta history](https://history.alberta.ca/energyheritage/sands/origins/the-fur-trade-and-albertas-oil-sands/peter-pond.aspx)

😆 

Happy to drop it down to 1750  instead, but my inking is to trust C&C on this one either way.

deployments/api/tests/db/test_source_priority_seed.py
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from stitch.api.db.model import OGFieldSourcePriority
+from stitch.ogsi.model import SOURCE_PRIORITY
+
+
+@pytest.mark.anyio
+async def test_seeded_priorities_match_source_priority(
```
AlexAxthelm, line 20: Thank you

AlexAxthelm (approved):

PR #194, opened 2026-07-24
Feat/rmi write (STIT-495)

AlexAxthelm:
Allow manual creation of RMI-type sources


mbarlow12 (commented):
Not necessarily blocking, but it seems a little cumbersome that we have to edit and save attributes individually instead of a more general "Edit" state where multiple attributes could be edited in one UI.

E.g. clicking "Edit" enters an "editing" state in the UI, user can perform any edits/overwrites (including re-prioritization), user needs to click "Save" to persist. Save would read the form data, and new data gets pushed as a new RMI source & re-prioritizations get separate API calls.

As I write this, it makes me think a `PUT` request on `/oil-gas-fields/{id}` might actually be appropriate. The API looks like `PUT`, but internally, we create the RMI source and persist reprioritizations.

mbarlow12 (approved):

AlexAxthelm:
@mbarlow12 https://rmi1.atlassian.net/browse/STIT-618 to capture multi-edit

PR #197, opened 2026-07-24
feat: add ETL call for `bc` scaffold `alb`

mbarlow12:
## What

- Platform-side plumbing for the new **BC** (BCER) and **Alberta** (AER) sources.
- Actual ETL changes are in [`stitch-etl-poc` PR](https://github.com/RMI/stitch-etl-poc/pull/17); this just makes the API + frontend aware of `bc`/`alb` so those posts are accepted and triggerable.

## Changes

- **OGSI model**: add `bc`/`alb` source keys + `BCSource`/`ALBSource` to the
  `OGFieldSource` union, and a canonical `SOURCE_PRIORITY` (single source of
  truth for the coalescer, query-param default, and DB seed).
- **Auth**: `source:read:bc` / `source:read:alb` permissions. **DOUBLE CHECK AUTH0 TENANT!!!**
- **Migration**: `add_bc_alb_source_priority` seeds the priority table from a "snapshot" (aka hardcoded copy) of `SOURCE_PRIORITY` so the end-state provably matches the canonical order.
- **Frontend**: wire the **BC** panel on the ETL dashboard to its actual start params (`payload_limit` + `use_cached`). Alberta keeps placeholder until its ETL lands.

## Verification
- local docker (AUTH_DISABLED=true) w/ existing records: 233 sources added (as expected from analysis of ARCgis data)

## Notes

- Depends on [`stitch-etl-poc#17`](https://github.com/RMI/stitch-etl-poc/pull/17) for anything to actually flow.
- Alberta will be wired in a subsequent PR
- consider `bcer` instead of `bc`?


deployments/stitch-frontend/src/constants/sourceMeta.js
AlexAxthelm: @mbarlow12 @jasonbrownrmi it seems like these `SOURCES` and `SOURCE_PRIORITY` variables are one of the big things that we could stop tracking manually if we pulled out a generated library of some kind?

But more immediate, I think the priority list on line 7 also needs updated?

AlexAxthelm (changes requested):
I think I see this one blocking, but not a big one.

Beyond that, I think that using `bcer` over `bc` makes sense. I generally prefer more explicit things if they're being used as FK values, since it reduces the chance of a conflict if we ever add some other `bc`-named organization.

AlexAxthelm (approved):

PR #210, opened 2026-07-31
STIT-611 Show resource list when fetching new data

jasonbrownrmi:
ResourcesTable returned null whenever its resources array was empty, so the whole table (including rows the user was just looking at) disappeared during every refetch triggered by an initial load, page change, filter, or sort, snapping back once data arrived.

- The resources table fully collapsed to nothing every time its query re-fetched — initial load, page change, filter change, or sort — because `ResourcesTable` returned `null` whenever `resources` was empty, and `resources` reset to `[]` while the query was in flight (`data` was `undefined` until the fetch resolved).
- Fix: added `placeholderData: keepPreviousData` (TanStack Query v5) to the resources list query in `queries/resources.js`, so `data` no longer resets to `undefined` mid-refetch — the previously loaded page's rows simply stay on screen for the duration of the new fetch.
- `ResourcesTable` now dims to 50% opacity and shows a centered spinner overlay while `isFetching` is true, with `aria-busy` on the table and an `sr-only` "Updating resources..." announcement for screen readers. The dimmed table is also non-interactive (`pointer-events-none`) for the duration of the fetch.
- `ResourcesView` keeps its plain "Loading resources..." text for the one case that still needs it: the very first load ever, before any data has arrived (there's nothing yet to dim).


mbarlow12 (approved):
LGTM

AlexAxthelm (approved):

PR #223, opened 2026-08-05
feat: optional `stake` and accurate seed data (STIT-528, STIT-562)

mbarlow12:
- adds owner/operator seed data
- makes stake for owners/operators optional, w/ tests & UI adjustments

**AI usage:** root cause investigation in `stitch-etl-poc`, CC implementation here was straightforward


deployments/seed/src/stitch/seed/payloads.py
```diff
@@ -63,11 +75,43 @@ def _year_triplet(rng: random.Random) -> tuple[int | None, int | None, int | Non
     return discovery, production_start, fid
 
 
+def _company(fake: Faker) -> str:
+    # Faker likes commas in company names; drop them to keep values simple.
+    return fake.company().replace(",", "")
```
AlexAxthelm, line 80: We should be able be able to handle commas in company names? seems like a blocker in that these should get escaped, not dropped.

deployments/seed/data/003-mock-data.json
```diff
@@ -102,8 +114,15 @@
           "last_updated": "2024-10-15T00:00:00Z",
           "name_local": null,
           "state_province": "Rogaland",
-          "region": "North Sea",
+          "region": "Russia and Caspian Sea",
           "basin": "North Sea",
+          "owners": [
+            {"name": "KazMunayGas", "stake": null},
```
AlexAxthelm, line 120: NB: Noting that the owners appear to be respecting the coalesce rules correctly (wm owners are presented rather than GEM ones), but these info cards aren't the same as the other fields (no click to see all values or edit)

AlexAxthelm (approved):
tentatively approved, pending discussion of commas in company names

PR #229, opened 2026-08-11
Refactor/stit 645 query layer factories

jasonbrownrmi:
## Summary

- Deletes the `resourceKeys` registry - key-builders are now private, used only inside `resourceQueries` factories (wrapped with TanStack's `queryOptions()`). Every consumer reads `.queryKey` off the same factory a hook uses, instead of a separately-maintained set of key builders.
- Drops `enabled=false` as the default on read hooks in `useResources.js`. Hooks now default to `enabled=true`, gated by a real check on their own args (`id != null`) instead of requiring every caller to remember to flip a flag.

## Why

`resourceKeys` living separately from its fetchers already caused one shipped bug (Clear Cache resetting the wrong query). This closes that class of bug structurally - a key and its fetcher can't drift apart when there's only one function producing both.

Found a live instance of the same bug during the refactor: `useResourceMock` built a `"detail"` key while its real counterpart used `"view"`. Fixed.

Stacked on `fix/STIT-446-save-cache-invalidation` (not yet merged); kept in sync with it.

## Test plan

- [x] `npx vitest run` — 286/286 passing
- [x] `npm run lint` — clean
- [x] `npx prettier --check` — clean
- [ ] Manual smoke test (blocked for me by Auth0 login, no dev bypass found)

Depends on: [STIT-446]
Closes [STIT-645]

[STIT-446]: https://rmi1.atlassian.net/browse/STIT-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployments/stitch-frontend/src/components/ResourceFieldCard.jsx
```diff
@@ -126,53 +126,34 @@ function MoveButtons({ onMoveUp, onMoveDown, canMoveUp, canMoveDown, label }) {
...
-      // Success unmounts this form (parent exits edit mode), so there is no
-      // `isSaving` to reset — only the error path below keeps the form mounted.
+      // Resolves after the mutation's cache invalidation, so the refreshed
+      // data is on its way before the form closes.
+      await createSource.mutateAsync({ resourceId, payload });
       onSaved();
-    } catch (err) {
-      setSaveError(err.message || "Failed to add value.");
-      setIsSaving(false);
+    } catch {
```
AlexAxthelm, line 154: NB: If this is an empty `catch`, do we need it here?

deployments/stitch-frontend/src/components/ResourceView.jsx
```diff
@@ -18,19 +18,21 @@ export default function ResourceView({
   const queryClient = useQueryClient();
   const [inputId, setInputId] = useState(1);
   const id = initialID ?? inputId;
+  // Embedded mode (showControls=false, given a known id) loads right away;
+  // interactive mode waits for the user to click Fetch / press Enter, which
+  // call refetch() directly below — refetch() always runs regardless of
+  // `enabled`, so those still work while this stays false.
+  const autoFetch = !showControls && initialID != null;
```
AlexAxthelm, line 25: Note from my Claude: nothing in `src/` renders this component — only its own test file and two README lines reference it.
```
git grep -n "ResourceView" -- deployments/stitch-frontend \
  | grep -v MergedResourceView | grep -v ResourcesView
```
Which means this `autoFetch` gate is the codebase's *only* remaining `enabled: false` — the exact pattern the ticket set out to remove — and it's defending a path the app never executes. It also means AC #4 ("embedded `showControls=false` mode fetches on its own") can't be validated by the smoke test, since there's no embedded instance.

deployments/stitch-frontend/src/components/ResourceFieldCard.jsx
```diff
@@ -302,10 +283,16 @@ function FieldSourcesPanel({
       // winner / collapsed value may have changed).
       await Promise.all([
         queryClient.invalidateQueries({
```
AlexAxthelm, line 285: My Claude flagged the issue below, but it might be hallucinating: I see the detail page update on both adding a source and reordering.

----

This is now the last hand-rolled write path: manual fetcher, manual two-key invalidation — while `AddSourceForm` two functions up goes through `useCreateSourceForResource` and invalidates `[endpoint]`.

The practical effect: reordering source priority changes the coalesced winning value, which shows in the resources list rows and in filter options. Those stay stale here, while *adding* a source refreshes them. That's the same drift the PR is closing, left in place one function away from the fix.

A `useUpdateFieldSourcePriority` hook alongside the other two would close it and drop the last `createAuthenticatedFetcher` call out of this component. Worth doing here rather than as a follow-up, since this PR is what makes the asymmetry visible.

deployments/stitch-frontend/src/hooks/useResources.js
```diff
@@ -327,12 +336,45 @@ function useMergeCandidatesMock(endpoint = "oil-gas-fields", enabled = false) {
...
+  const config = useConfig();
+  const queryClient = useQueryClient();
+  return useAuthenticatedMutation({
+    mutationFn: (fetcher, { resourceId, payload }) =>
+      createSourceForResource(config, resourceId, payload, fetcher, endpoint),
+    // Every resourceQueries key is prefixed by [endpoint] (enforced by
+    // resources.test.js), so this one-segment key invalidates everything
+    // cached for the endpoint — lists, details, views, field-sources,
+    // filter-options, and merge-candidates.
+    onSuccess: () => queryClient.invalidateQueries({ queryKey: [endpoint] }),
```
AlexAxthelm, line 367: again, from Claude: The `[endpoint]` prefix invariant is currently enforced by a comment and a guard test, but not by the code — `keys.all()` is private to `resources.js`, so this can't call it and has to hardcode the literal instead.

Small change that makes it structural:

```
// resources.js
export const resourceQueries = {
  all: (endpoint = "resources") => keys.all(endpoint),
  // ...
};

// here
onSuccess: () =>
  queryClient.invalidateQueries({ queryKey: resourceQueries.all(endpoint) }),
```

Same for line 377. The `it.each` prefix test then becomes belt-and-braces instead of the only thing standing between us and a silently-missed invalidation — which is exactly the argument this PR makes for killing `resourceKeys` in the first place.

AlexAxthelm (approved):
Some manual testing in the PR preview looks good to me, but passing the PR past my Claude flagged a few notes (nothing I see blocking)

PR #230, opened 2026-08-12
feat(client): add methods to bulk access oil-gas-field-sources

AlexAxthelm:
Contributes to https://rmi1.atlassian.net/browse/STIT-529

Support [etl-poc PR](https://github.com/RMI/stitch-etl-poc/pull/23) to deduplicate before POST

----

Contributes to [STIT-529](https://rmi1.atlassian.net/browse/STIT-529). Supports [stitch-etl-poc#23](https://github.com/RMI/stitch-etl-poc/pull/23), which needs to read existing sources back before POSTing so a second ETL run doesn't duplicate them.

## What changed

**`AsyncStitchClient` gains two source-listing methods** (`9a8965e`) — `list_oil_gas_field_sources_page` and `iter_oil_gas_field_sources`, mirroring the existing `*_oil_gas_fields` pair. Unlike the resource route, each item is a flat source view (every business field plus `id` and `source`), which is what makes a one-pass fingerprint index possible.

Three things in there are load-bearing rather than stylistic:

- **`sort_by=id` is pinned internally, not exposed.** The route defaults to name-ascending, under which a concurrent insert lands *behind* the paging cursor and shifts every later page — silently skipping rows during a full scan. Sorting on the source id makes new rows append past the last page.
- **An empty `source` sequence raises.** httpx drops it from the query string entirely, which the API reads as "no filter" — i.e. all seven sources, the opposite of the intent.
- **Partial `source:read:*` grants return 200 with rows silently dropped**, not 403. Documented on the method so callers don't treat an empty result as authoritative.

**Client-side pagination guard** (`d37583f`) — `page_size` is bounded 1..200 (the server's `PaginationParams` cap) and `page >= 1`, on both routes, failing before the request rather than on a server 422. `MAX_PAGE_SIZE` is now exported from `stitch.client` so callers sizing a full scan don't reach into `async_client`.

**Bare-string `source` guard** (`0438660`) — `str` satisfies `Sequence`, so `source="gem"` expanded to `source=g&source=e&source=m`; the API rejected those as invalid enum values and the 422 surfaced far from the typo. Now a `TypeError` naming the fix. Caught by Copilot review.

**Unrelated:** `68558ca` drops the CPU/memory pin on the entity-linkage deploy job now that it streams via `iter_oil_gas_fields` rather than materializing the full set.

## AI assistance

Claude drafted the two client methods, the pagination and string guards, and all 13 new tests, working from the existing `list_oil_gas_fields_page` / `iter_oil_gas_fields` implementations. The `sort_by=id` decision came out of reading the API's `_build_sort_clauses` to check what offset pagination actually does under a concurrent write.

## Tests

`packages/stitch-client/tests/test_async_client.py`, 46 → 55 cases. New coverage: request shape including the pinned sort params; repeated `source` params reaching the wire (asserted on the raw query string, since `dict(url.params)` collapses repeats); `source` forwarded on page 2+, not just the first request; empty `source` and bare-string `source` rejected before any request; `page_size` bounds on both routes and through the generator; multi-page streaming, empty first page, short-page termination; 403 raising `StitchAPIError` with `status_code`, and propagating out of the async generator rather than yielding empty.

`make py-test` — 543 passed. `make py-lint` / `make py-format-check` clean.

## Follow-up

The four-condition pagination termination loop is now duplicated across three methods (`iter_oil_gas_fields`, `collect_oil_gas_fields`, `iter_oil_gas_field_sources`). Copied verbatim rather than refactored here to keep this PR to one concern — tech-debt ticket to follow.

[STIT-529]: https://rmi1.atlassian.net/browse/STIT-529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/stitch-client/src/stitch/client/async_client.py
+++ b/packages/stitch-client/src/stitch/client/async_client.py
@@ -268,6 +268,7 @@ class AsyncStitchClient:
             # which a concurrent insert can land behind the paging cursor and
             # shift every later page, silently skipping rows. Sorting on the
             # source id makes new rows append past the last page instead.
+            # As such, `sort_by` and `sort_order` are deliberately NOT overrideable
             "sort_by": "id",
             "sort_order": "asc",
         }
--- a/packages/stitch-client/src/stitch/client/async_client.py
+++ b/packages/stitch-client/src/stitch/client/async_client.py
@@ -254,6 +254,7 @@ class AsyncStitchClient:
                 dropped -- only a token holding *none* of them gets a 403.
 
         Raises:
+            TypeError: if ``source`` is a bare string rather than a sequence.
             ValueError: if ``page``/``page_size`` are out of range, or ``source``
                 is an empty sequence.
             StitchAPIError: on a non-2xx response (403 when the token holds no
@@ -271,6 +272,15 @@ class AsyncStitchClient:
             "sort_order": "asc",
         }
         if source is not None:
+            if isinstance(source, (str, bytes)):
+                # A bare string is itself a Sequence, so list("gem") would
+                # expand to ["g", "e", "m"] and the request would go out as
+                # source=g&source=e&source=m. The API rejects those as invalid
+                # enum values, so the 422 surfaces far from the actual typo.
+                raise TypeError(
+                    "source must be a sequence of source keys, not a bare "
+                    f"{type(source).__name__} (did you mean [{source!r}]?)"
+                )
             source_keys = list(source)
             if not source_keys:
                 # httpx drops an empty sequence from the query string entirely,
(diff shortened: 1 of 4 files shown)


packages/stitch-client/src/stitch/client/async_client.py
```diff
@@ -217,6 +227,123 @@ async def collect_oil_gas_fields(
...
+        """
+        self._validate_page_params(page, page_size)
+        params: dict[str, Any] = {
+            "page": page,
+            "page_size": page_size,
+            # Stable forward scan. The route defaults to name-ascending, under
+            # which a concurrent insert can land behind the paging cursor and
+            # shift every later page, silently skipping rows. Sorting on the
+            # source id makes new rows append past the last page instead.
+            "sort_by": "id",
```
jdhoffa:
```suggestion
            # source id makes new rows append past the last page instead.
            # As such, `sort_by` and `sort_order` are deliberately NOT overrideable
            "sort_by": "id",
```

I know the docstring above already suggests this, but it's worth just being explicit.

jdhoffa (approved):
> The four-condition pagination termination loop is now duplicated across three methods (iter_oil_gas_fields, collect_oil_gas_fields, iter_oil_gas_field_sources). Copied verbatim rather than refactored here to keep this PR to one concern — tech-debt ticket to follow.

Flagged under Tech Debt in this ticket for posterity: https://rmi1.atlassian.net/browse/STIT-652

Minor comment on documentation just to be super explicit, but otherwise this LGTM

PR #234, opened 2026-08-13
ci(hostnames): Update behavior for Azure SWA deployments

AlexAxthelm:
Two changes to how the CD pipeline targets Azure Static Web Apps, plus the docs that explain them.

**`main` now deploys to the dev SWA's production environment.** `production-branch` was hardcoded to `production`, so pushes to `main` landed in a *named preview* environment (`witty-mushroom-…-main.westus2.1.azurestaticapps.net`). It is now lane-aware — `main` for the development lane, `production` otherwise. Azure only allows custom domains on a site's production environment, so this is the prerequisite for `stitch-dev.rmi.org`.

**Scale-to-zero policy is now explicit, and narrower than before.** `min-replicas` was keyed off lane (`!= 'development'`). `resolve-context` now decides it once and exposes it as an `always-on` output:

| Deployment | `always-on` |
| --- | --- |
| push to `production` | `1` |
| push to `main` | *(empty)* |
| `staging` lane PRs (into `production`, or from `demo/*`) | *(empty)* |
| `development` lane PRs | *(empty)* |

⚠️ **This is a deliberate reduction.** Per the cost decision with @jdhoffa, only the production release keeps a warm replica — so `staging`-lane deployments lose the one they have today, and `demo/*` branches and PRs into `production` will now cold-start. The offset is the login-page prewarm follow-up below, not a wider policy.

## Bug this fixes

`README.md` advertises the dev lane as the **root** `witty-mushroom-….azurestaticapps.net` URL, but the pipeline handed the API `…-main.westus2.1.azurestaticapps.net` as its single allowed CORS origin. Anyone using the advertised URL got CORS failures on every API call. That root environment also hadn't been deployed to since 2026-06-01, so it was serving a stale build.

The frontend-origin computation is now `PR → preview template, anything else → FRONTEND_PRODUCTION_URL`. That's true for all three lanes, so the `dress-rehearsal` special case goes away.

## Verification

The resolve script was extracted and run against every event shape (pushes to `main`/`production`/an unsupported ref, PRs into `main`/`next`/`production`, and `demo/*`). Lane and deployment-name outputs are byte-identical to `main`; the only delta is `always-on`. Generated names matched live Azure resources (`pr-0224-next-c6c1`, `pr-0232-demo-dedupe-7c63`), which is where the corrected doc examples come from.

## Docs

`CI_DEPLOYMENTS.md` gains a **Custom domains** section (hostname↔lane table, DNS ownership, add-a-hostname procedure) and a rewritten scale-to-zero section. It also picks up two corrections:

- The branch→lane table was stale — it claimed `deployment_name=production` and `pr-<number>` where the code produces `dress-rehearsal` and `pr-0042`.
- The CPU/memory section still said `entity-linkage` and `etl` both pin `cpu: "2.0"` / `memory: "4.0Gi"`. #230 unpinned `entity-linkage` without updating the docs, so merging `main` here reconciles them.

`stitch-frontend/README.md` gets the Auth0 guidance: the SPA derives `redirect_uri` from `window.location.origin` and has no callback route, so every hostname needs its own **bare origin** entry — no trailing slash, no path.

## Follow-up (not in this PR)

Nothing user-visible changes until step 3. Merging this is safe on its own.

Only `stitch-dev.rmi.org` is being assigned now. `stitch.rmi.org` is held back until a real production SWA exists, so the name never has to move between resources — it is deliberately **not** parked on `stitch-staging`, which serves the dress-rehearsal lane. (Moving a bound domain between two SWAs in the same slice requires downtime.)

1. **Set up the domain name on the SWA** — RMI IT adds the CNAME (`rmi.org` is hosted at Network Solutions; not self-service), then bind it in Azure and let the managed cert issue: `stitch-dev.rmi.org` → `witty-mushroom-017a3dc1e.1.azurestaticapps.net`.
2. **Set it as the default domain**, so Azure 301-redirects the old `azurestaticapps.net` hostname. ⚠️ Verify an open PR's preview URL still loads afterward — Azure's docs say a default domain redirects "requests to other domains" without saying whether preview hostnames count.
3. **Flip `FRONTEND_PRODUCTION_URL`** on the `development` GitHub Environment, and update the `README.md` links. This is the cutover.

Auth0 registration isn't needed for `stitch-dev.rmi.org` — the `development` lane runs `AUTH_DISABLED=true`. It will be required for the production hostname.

## Out of scope

- **Login-page API prewarm** — separate ticket; mitigates the cold starts this policy accepts. (included in #238)
- **Per-PR RMI hostnames** (STIT-615 AC). Not possible with SWA custom domains at all — preview environments can't hold one. Needs Front Door or one SWA per lane.
- **`main`'s database is dropped on every push** ([`deploy-db.yml:106`](.github/workflows/deploy-db.yml#L106)). Worth revisiting once `stitch-dev.rmi.org` is a shareable URL, but not a hostname change.
- The `development` lane runs `AUTH_DISABLED=true`.

Solves [STIT-625] — scope changed during planning: `stitch-staging.rmi.org` was dropped in favour of `stitch-dev.rmi.org` now and `stitch.rmi.org` once a production SWA exists. The ticket's summary and AC still need updating to match.

[STIT-625]: https://rmi1.atlassian.net/browse/STIT-625


deployments/CI_DEPLOYMENTS.md
```diff
@@ -2,25 +2,31 @@
...
 - `deployment_name`: concrete runtime target name used for DB and app naming
+- `always_on`: whether this deployment keeps a warm Container App replica
 
 Branch behavior is:
 
 - push to `main` -> `deployment_lane=development`, `deployment_name=main`
-- any PR not targeting `production` -> `deployment_lane=development`, `deployment_name=pr-<number>`
-- push to `production` -> `deployment_lane=dress-rehearsal`, `deployment_name=production`
+- any PR not targeting `production` -> `deployment_lane=development`, `deployment_name=pr-<zero-padded number>`
+- push to `production` -> `deployment_lane=dress-rehearsal`, `deployment_name=dress-rehearsal`
```
AlexAxthelm, line 16: @jdhoffa I'm guessing that we don't want to point `stitch.rmi.org` to dress rehearsal (we should wait to assign that until we've actually stood up a prod SWA)?

AlexAxthelm:
Decisions with @jdhoffa in meeting:
* deploy `stitch-dev.rmi.org` now, assign prod when we create the actual SWA.
* Costs: Set prod to stay awake (always-on 1), all others can fall asleep for now.
* Followup: send request to api from logion page so that it can start waking up while user logs in.

jdhoffa:
> Decisions with @jdhoffa in meeting:
> 
> * deploy `stitch-dev.rmi.org` now, assign prod when we create the actual SWA.
> * Costs: Set prod to stay awake (always-on 1), all others can fall asleep for now.
> * Followup: send request to api from logion page so that it can start waking up while user logs in.

PR body here is looking stale, can you please update it so it's in line with what the PR actually does for posterity?

jdhoffa (approved):

PR #235, opened 2026-08-13
feat(api): make batch traffic yield to interactive users (STIT-644)

AlexAxthelm:
tl;dr: add an optional HTTP header `X-Stitch-Traffic-Class: batch` to allow requests to flag themselves as lower priority (They will wait until untagged requests are done before processing)

Note this doesn't actually solve the problem of "EL makes a lot of traffic", but it does help control by putting up a yield sign for the batch jobs. This is a low-effort very basic approach, but it doesn't exclude any better options later. We should consider adding to ETL and seed at some point.

----

## Problem

[STIT-644] — a linkage run saturates the shared API and leaves too little capacity for a human to use the app (found in the B&L demo).

Entity linkage calls the same public API surface as the frontend. That's a good programming model and this PR keeps it.

## Why not a concurrency limit

EL's in-flight HTTP concurrency is **already 1**: `JobManager` permits one run and `link_all` awaits every call sequentially. A semaphore sized above 1 would never engage.

EL saturates through **duty cycle**, not parallelism — ~3+ requests per resource, back to back, against the most expensive query in the system, with zero think time. So this is a rate control, not a concurrency limit.

## What this does

A request tagged `X-Stitch-Traffic-Class: batch` waits while interactive requests are in flight or recently finished, then is **admitted anyway** after `max_wait` — never a 429, so it cannot fail a linkage run. Untagged requests are interactive, so there is nothing worth spoofing.

- Pure ASGI (not `BaseHTTPMiddleware`, whose `call_next` returns at response *headers* and would release the in-flight count too early), registered innermost so a deferred request resolves no dependencies — no token verify, no DB session.
- Health paths skip the gate **and** the activity accounting; otherwise the every-5s healthcheck would make any `quiet_ms > 5000` starve batch traffic with zero human load.
- Off by default (`BATCH_YIELD_ENABLED`); warns and stays inert in prod.

Two per-request cost reductions that help interactive traffic equally:
- Validated token claims cached (keyed on a SHA-256 digest, honoring the token's own `exp`, failures never cached) — removes an RSA verify from every repeat request.
- DB pool limits stated explicitly. Same values as SQLAlchemy's defaults, so no behavior change; the point is the 15-connection ceiling is now visible.

Enable locally with `BATCH_YIELD_ENABLED=true` in `.env`.

## Scope

Not the root cause: `link_all` is O(resources × pages) against the `q=` search, which **STIT-619** tracks. This makes the shared server usable meanwhile. The two compose — the gate acts per request at admission, so it still works once STIT-619 gives EL concurrent requests.

Deriving traffic class from caller identity rather than a header becomes possible with M2M (**STIT-469** / **STIT-424**); the header stays the mechanism either way.

The entity-linkage container cpu/memory unpin is on a separate branch.

## AI assistance

Claude Code was used for codebase exploration, design review, and drafting the implementation and tests.

**Verified locally:** `make py-test` (562 pass), `py-lint`, `py-format-check`, `py-lock-check` all green.

**Tests:** `deployments/api/tests/test_admission.py` (40 tests, 0.2s) covers classification, exemptions, quiet-window behavior via an injected clock, in-flight accounting under handler exceptions and client cancellation, registration order, and prod inertness — plus one deliberate real-clock test so a mis-wired default clock/sleep can't slip through. Claims-cache behavior in `test_auth_unit.py`.

**Not verified:** the end-to-end docker comparison (baseline vs. under-linkage latency via `tools/analyze_logs.py --group-by scenario`). Worth doing against the preview env.

[STIT-644]: https://rmi1.atlassian.net/browse/STIT-644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployments/api/src/stitch/api/db/config.py
```diff
@@ -54,6 +54,18 @@ def get_engine() -> AsyncEngine:
         settings.get_database_url(),
         echo=not settings.is_prod,
         pool_pre_ping=True,
+        # Stated explicitly rather than inherited. These are SQLAlchemy's own
+        # defaults, so this changes no behavior — the point is that the ceiling is
+        # now visible: 15 connections per process, and a 16th request waits up to
+        # pool_timeout before failing. Worth knowing because every authenticated
+        # request checks out *twice* (get_current_user opens its own session before
+        # the handler's UnitOfWork), and because each replica brings its own pool,
+        # so the real cap against Postgres is 15 x replicas.
```
mbarlow12: We need to get the agent output in line. I think this kind of commentary actually contributes to a positive feedback cycle where Claude/Codex are more likely to keep going down the same path instead of questioning previous assumptions.

- "the ceiling is now visible"
- where does the "15 connections per process" come from??? I don't doubt it's there somewhere in the SQLAlchemy docs, but that seems pretty important to reference.
- "every authenticated request checks out *twice*" could be "every request opens 2 sessions"...and have we independently confirmed that this isn't a hallucination? Or that the first session closes before the second opens?
- "the real cap against Postgres is 15 x replicas"

This all for a change that ostensibly "changes no behavior".

deployments/api/src/stitch/api/admission.py
```diff
@@ -0,0 +1,207 @@
+"""Cooperative admission control: batch traffic yields to interactive traffic.
+
+Entity-linkage calls the same public API surface as the frontend. That is a good
+programming model, but it means a linkage run can crowd humans off a shared dev
+server. It does so through *duty cycle*, not parallelism: the bulk pass is
```
mbarlow12: > That is a good programming model

Is it, though?

Similar comment as above, but this sort of text further pushes agents toward justifying whatever was done, regardless of whether the decisions & designs are actually beneficial to us.

deployments/api/src/stitch/api/admission.py
```diff
@@ -0,0 +1,207 @@
...
+# Matched against the raw ``scope["path"]``, which is the full path today because
+# nothing fronts the API with prefix stripping and uvicorn runs without
+# ``--root-path``. Revisit if either changes.
+_HEALTH_PATH = "/api/v1/health"
+
+# How precisely a waiting request notices the quiet window clear. Deliberately
+# not a setting: it is not a delay we impose (the quiet check runs *before* the
+# first sleep, so an idle server adds zero latency), and exposing it invites
+# reading it as one. Must stay well below ``quiet_s`` or it dominates the
+# resume time as overshoot.
```
mbarlow12: I have no idea what this means.

mbarlow12 (commented):
I'll be honest, I didn't review the whole thing. I got through a handful of the comments, and started getting pretty frustrated by what appears to be AI-generated, confusing text. I've left some more specific comments.

From a really broad perspective, I don't think I can sign off on this as I don't think it's an appropriate approach. We're trying to navigate competing request traffic using internal queueing and request headers.

From the PR body:
> Why not a concurrency limit
> EL's in-flight HTTP concurrency is already 1: JobManager permits one run and link_all awaits every call sequentially. A semaphore sized above 1 would never engage.

> EL saturates through duty cycle, not parallelism — ~3+ requests per resource, back to back, against the most expensive query in the system, with zero think time. So this is a rate control, not a concurrency limit.

I'd want to see clearer justification here. It seems like there's some confusion about where the bottleneck is. If EL's API is saturated, that doesn't touch the main `stitch-api`, and if the `stitch-api` is "saturated" (again, I'm skeptical of this claim) then fielding requests in parallel is absolutely a potential solution.

I think STIT-644 needs more thorough root cause investigation. If that was done, can we share that somewhere? If the observed behavior was a laggy/slow UI and that lag was caused by too many blocking/synchronous requests, that, to me, points to addressing why they're blocking for so long or why there are so many of them or why they should to compete with user traffic at all.

Would pulling the entity-linkage logic into `stitch-api` have been a larger change?
Would a reverse proxy be a viable consideration to replace `uvicorn` to get process level concurrency across requests? (not advocating for this given the complexity it adds)
Would increasing the response size from stitch-api to EL allow EL to work independently for longer, thereby freeing the API?

mbarlow12:
Also, this has highlighted some of my own more cavalier approach to letting AI run & write more freely. I think we probably need to start instituting policies at the SPD & stitch level.

AlexAxthelm:
Closing this, but may want to revisit the concept in the future

PR #258, opened 2026-08-28
feat: redirect merged-away resources to their merged target (STIT-418)

AlexAxthelm:
Enables automatically returning the repointed-to resource (or last along a repointing chain in the case of multiple merges), rather than the empty (null-shell) resource. Includes a new `requested_resource_id` field to inform users that the resource they are seeing is technically different than the one they attempted to access directly.

Generally moves all read-path operations to `get_resolved` instead of the raw `get` operation.

Also includes an update to detail page to let users know in the UI if they click on an outdated link (such as this one: https://witty-mushroom-017a3dc1e-258.westus2.1.azurestaticapps.net/oil-gas-fields/15, which now takes you to resource `38`)

----

## STIT-418 (part 1 of 2) — redirect on the resource detail path

Merging in Stitch is *repointing*: merging A and B creates a new resource C and
sets `repointed_id` on A and B. Previously a request for a merged-away resource
returned HTTP 200 with an all-null "null shell", so `/oil-gas-fields/A` rendered
a blank-but-successful page.

This makes any request that resolves to a repointed resource return the resource
it was merged into:

- **API:** `GET /oil-gas-fields/{id}` and `/{id}/detail` follow the repoint chain
  to the terminal resource and return its data, with a new nullable
  `requested_resource_id` (the originally-requested id, or null when you got
  exactly what you asked for). Additive to the schema.
- **Frontend:** `ResourceDetailPage` redirects to the canonical URL
  (`navigate(..., { replace: true })`) when the returned id differs from the URL.

Resolution lives in a read-only helper (`get_resolved`); the write-path `get`
still returns the exact requested row, since merges must operate on it.

**Scope:** part 1 of 2. Part 2 (Merge Review pane, AC 3) is the stacked
follow-up — do not close STIT-418 when this merges.

### Tests
Backend: multi-hop chain collapse, non-repointed self-resolve, 404; route
integration for the resolved id + `requested_resource_id` + real data; OpenAPI
contract for the new field. Frontend: redirect fires once with `replace:true`;
no redirect for a live resource; no loop.

### AI assistance
Implemented with Claude Code (code + tests). Verified: API pytest + frontend
vitest suites pass locally.

### DIFF
--- a/deployments/stitch-frontend/src/pages/ResourceDetailPage.jsx
+++ b/deployments/stitch-frontend/src/pages/ResourceDetailPage.jsx
@@ -1,6 +1,6 @@
 import { useEffect, useId, useState } from "react";
 import { useAuth0 } from "@auth0/auth0-react";
-import { useParams, useNavigate } from "react-router";
+import { useParams, useNavigate, useLocation } from "react-router";
 import {
   useCreateSourceForResource,
   useResourceDetail,
@@ -554,9 +554,15 @@ function SourcesSection({ sources }) {
 export default function ResourceDetailPage() {
   const { id } = useParams();
   const navigate = useNavigate();
+  const location = useLocation();
   const numericId = Number(id);
   const validId = Number.isFinite(numericId);
   const endpoint = "oil-gas-fields";
+  // STIT-418: set by the redirect below when a merged-away id is bounced to its
+  // canonical resource. Ephemeral by design — it lives in router state, so a
+  // fresh load, refresh, or new tab of this URL carries none and shows no
+  // notice; it appears only when the user actually got redirected here.
+  const redirectedFrom = location.state?.redirectedFrom ?? null;
   // Enabled (not manually refetched) so that save mutations' cache
   // invalidations refetch it — invalidation never refetches disabled queries.
   const {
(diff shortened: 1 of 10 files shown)


deployments/stitch-frontend/src/pages/ResourceDetailPage.jsx
```diff
@@ -567,12 +573,39 @@ export default function ResourceDetailPage() {
...
       <Button onClick={() => navigate(-1)} variant="ghost" className="mb-6">
         ← Back
       </Button>
 
+      {redirectedFrom != null && (
+        <div className="mb-6 rounded-md border border-line bg-surface px-4 py-3 text-sm text-ink">
+          Resource <span className="font-medium">{redirectedFrom}</span> was
+          merged into this record, so you were brought here.
+        </div>
+      )}
```
jdhoffa: > Includes a new requested_resource_id field to inform users that the resource they are seeing is technically different than the one they attempted to access directly.

NB: Would be nice if this could be a toast notification or something else short-lived. 
Captured on backlog here: https://rmi1.atlassian.net/browse/STIT-720

jdhoffa (approved):
Looks good!
A NB comment for follow-up work in a future feature.

PR #259, opened 2026-08-28
Feat/merge review staleness

AlexAxthelm:
Extends #258 to address the merge-review dead end: when two pending merge
candidates share a source resource, approving one merges that resource away
(its original becomes an all-null shell), leaving the other candidate
functionally invalid. This surfaces that state — hiding the approve/deny
controls and telling the reviewer which resource was merged into which.

Added as a separate PR (stack) primarily because this one has some choices around the ergonomics of merge review that we may want to discuss more without blocking #258

----

## STIT-418 (part 2 of 2) — resolve repointed resources in the Merge Review pane

**Stacked on #258** (base branch `feat/redirect-detail`); review/merge part 1
first, then this retargets to `main`.

When an overlapping candidate is approved, a member of another pending candidate
gets merged away. Previously that candidate's Approve was guaranteed to 400 with
a message interpolating `repr()` on ORM objects (`Repointed: [<...object at
0x...>]`), and the only exit was Deny — recording a judgment no one made.

- **API:** merge-candidate views carry `repointed_resources` (which members moved
  and their terminal resource), computed for PENDING candidates via a single
  batched root lookup. The approve-time guard now names the terminal resource
  ("resource 102 is now resource 301") on both the merge-candidate and
  write-path guards.
- **Frontend:** the review pane shows a warning banner naming each move (linked),
  hides Approve/Deny when a member has moved, and annotates moved members with
  "(now 301)".

Also includes a small perf commit dropping redundant queries on the resolve
paths (`get_resolved` builds the entity from the row it already loaded; `create`
skips an always-empty staleness lookup).

The retire/requeue/`SUPERSEDED`-status flow is intentionally deferred until the
stale-candidate dead end actually blocks a reviewer.

### Tests
Backend: origin-column batch mapping, N+1 batch guard, PENDING-only rule, the
cross-candidate case end-to-end (approve 400s naming both ids, no `object at
0x`, stays PENDING), and the apply_resource_merge guard case. Frontend: stale
banner + hidden controls, facts annotation, compat guard, and a pure-function
suite for the staleness util.

### AI assistance
Implemented with Claude Code. Verified: API pytest + frontend vitest green.

### DIFF
--- a/deployments/api/src/stitch/api/db/merge_candidate_actions.py
+++ b/deployments/api/src/stitch/api/db/merge_candidate_actions.py
@@ -345,7 +345,10 @@ async def create_merge_candidate(
     )
     await session.flush()
     await session.refresh(candidate, ["items"])
-    return await _candidate_to_view_resolved(session, candidate)
+    # A newly created candidate's members were just validated as unrepointed, so
+    # nothing is stale yet -- pass an empty resolution instead of running the
+    # (always-empty) root lookup.
+    return _candidate_to_view(candidate, [])
 
 
 async def approve_merge_candidate(
--- a/deployments/api/src/stitch/api/db/merge_candidate_actions.py
+++ b/deployments/api/src/stitch/api/db/merge_candidate_actions.py
@@ -22,6 +22,7 @@ from stitch.api.entities import (
     MergeCandidateReviewRequest,
     MergeCandidateStatus,
     MergeCandidateView,
+    RepointedResourceView,
 )
 from stitch.ogsi.model import OGFieldSource
 from stitch.ogsi.model.og_field import OilGasFieldBase
@@ -33,7 +34,7 @@ from .model import (
     OGFieldSourcePriority,
     ResourceModel,
 )
-from .og_field_resource_actions import apply_resource_merge
+from .og_field_resource_actions import apply_resource_merge, repointed_merge_error
 from .utils import coalesce_resources_with_sources
 
 
@@ -62,11 +63,7 @@ async def _load_mergeable_resources(
 
     repointed = [r for r in results if r.repointed_id is not None]
     if repointed:
-        reprs = map(repr, repointed)
-        msg = f"Repointed: [{','.join(reprs)}]"
-        raise ResourceIntegrityError(
-            f"Cannot merge any resource that has already been merged. {msg}"
-        )
+        raise ResourceIntegrityError(await repointed_merge_error(session, repointed))
 
     return results
 
@@ -75,7 +72,13 @@ def _fingerprint(resource_ids: Sequence[int]) -> str:
     return ":".join(map(str, sorted(set(resource_ids))))
 
 
-def _candidate_to_view(model: MergeCandidateModel) -> MergeCandidateView:
+def _candidate_to_view(
+    model: MergeCandidateModel,
+    repointed_resources: Sequence[RepointedResourceView],
+) -> MergeCandidateView:
+    # `repointed_resources` is required so no caller can silently emit an empty
+    # staleness signal; resolve it with `_resolve_candidates` (batched) or
+    # `_candidate_to_view_resolved` (single).
     return MergeCandidateView(
         id=model.id,
         resource_ids=[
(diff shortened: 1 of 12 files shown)


mbarlow12 (commented):
Nothing here is badly executed. The problem is that these changes (and somewhat those in #258) work around existing design issues that should be addressed. I'm concerned merging it further reinforces those designs and will make future change harder.

---

## 1. AC discrepancy

STIT-418 AC 3: *"Same for any relevant Merge Candidates in the Merge Review pane."*
The AC asks merge candidates to **re-route** repointed resources, but these changes appear to only add metadata to the response to inform the reviewer of repointing.

A narrower reading of the AC affords a much smaller change. When approving (A,B) → D, update any pending candidate containing B to contain D instead:
- (B,C) becomes (D,C), stays PENDING, reviewer continues
- no new API field, no frontend util, no CTE change
- ~15 lines in `approve_merge_candidate`, same transaction

This might also warrant dropping the fingerprint in favor of a constraint that
a given resource id can't participate in more than one candidate at a time.

## 2. Additional code churn

Across #258 + #259, we've added ~1000 lines and two new permanent public API fields.

- `_parent_tree_cte` was stable for months; it's now rewritten to carry an origin
  column that serves one view field
- `get` is largely duplicated in `get_resolved`
- `requested_resource_id` (#258) has **zero** frontend consumers. `ResourceDetailPage` reads `location.state.redirectedFrom`, which it sets itself
- `repointed_resources` (#259) is a second public field for the same concept

Given how we do matching right now, have we actually observed overlapping ids in merge candidates? It seems like the edge case we're handling here is really, really unlikely.

## 3. Benefits of halting and addressing tech debt

I think it's worth making a decision on STIT-489: **all API endpoints live behind `stitch-api`**, keeping in mind that nothing has to move at once.

The point of committing to the direction is that it decouples the API contract from the implementation. Once an endpoint sits behind `stitch-api`, whether the work happens in a module, a proxied service, or something else is an implementation detail we can revisit later without affecting a consumer. This makes decisions that affect backend internals more reversible and flexible, and it's what makes the two changes below safe to pursue incrementally.

**Change 1: one pending candidate per resource.** A partial unique index on
`merge_candidate_items.resource_id` where the candidate is pending. If two pending candidates *cannot* share a resource, then staleness becomes impossible. I'm also pretty sure this removes the need for `fingerprint` as well.

**Change 2: entity-linkage becomes a module rather than a deployment.** Matching moves into `stitch-api` and excludes repointed resources and resources already claimed by a pending candidate. This would retain the ~220 lines of matching logic and drop essentially everything else from `entity-linkage`.

This also narrows matching to a single function with a stable signature, roughly `propose_merge_groups(session, available_ids) -> list[set[resource_id], ...]`. It's in-process with direct DB access, and we'd be free to update the implementation however we want as long as the signature is maintained.

Another way of thinking of this relates to STIT-672 (manual merge candidate creation): a "create merge candidates" form with linkage as an auto-fill feature (like AI enrichment). Proposals are ephemeral and generated on the fly, and we only persist whatever the user actually submits.

**Change 3: making the API more strictly RESTful (somewhat out of scope here).** This would help clarify the order of operations in this workflow. Candidates become a resource (in the REST sense) with the desired CRUD operations. Today

- `entity-linkage` creates and persists candidates
- `approve` both transitions the candidate and creates a resource
- merging is unreachable except through a candidate.

A more RESTful pattern would be:

- `/og-merge-candidates`: GET, POST
- `/og-merge-candidates/{id}`: GET (optionally PUT & DELETE)
- `/og-merge-candidates/proposals`: what "entity-linkage" becomes
  - ephemeral GET request that essentially pre-populates a form
  - clicking "Submit" is what persists the merge candidates (i.e. POST `/og-merge-candidates`)
- `/oil-gas-fields/merge`: approving a merge candidate sends a POST, guarded by checking the merge candidate table
  - could also be a POST to `/oil-gas-fields` with the constituent ids

This is a bigger conversation than this PR, but the point is that

- API consolidation allows us freedom to make such decisions
- stricter REST design makes it easier to reason about what should happen for a given operation.

Related open tickets:

| | |
| --- | --- |
| STIT-489 | Consolidate API routes in `stitch-api` |
| STIT-757 | 15-hour bulk run on prod volumes |
| STIT-644 | Entity Linkage slows down entire app |
| STIT-619 | Speed up the bulk pass (concurrency + read cache) |
| STIT-719 | Bulk run aborts on one transient HTTP timeout |
| STIT-621 | Durable, multi-worker job state |
| STIT-740 | Status stuck "Running" with no progress |
| STIT-620 | Bound the bulk-run result summary |

Setting STIT-489 aside, most of these exist because of the network boundary and in-memory job manager and how the current design/topology constrains our options. Many become either irrelevant or at least much easier to solve with direct DB access and focusing on application boundaries first.

jdhoffa:
@mbarlow12 and @AlexAxthelm I have a few thoughts here: 

1. We have 7 funded days left on Stitch, and no further funding yet in the hopper
2. This seems like a P0 fix (not merging in _something_ here will lead to un-defined user-facing behaviour)
3. "Working software over comprehensive documentation"

With that in mind, my inclination is to get this in, as a larger job-infrastructure refactor is out of scope in the given time-line, and we aren't in a position to **not** fix this before hitting production. 

Do you both agree or am I missing anything?

AlexAxthelm:
Superseded by #287

